PR #9, opened 2025-03-05
[PCC-73] Fix second-bot hang on REST-invoked bots

cbrianhill:
There are two changes here:
- Remove the `ProcessPoolExecutor` from the base image.  I found some evidence that this doesn't work super well with uvicorn, and we can live without it.
- Tell the `PipelineRunner` not to handle `SIGINT`.  After I fixed the first part, the bot would hang at shutdown due to this.

### DIFF
--- a/base/app.py
+++ b/base/app.py
@@ -2,7 +2,6 @@
 import asyncio
 import multiprocessing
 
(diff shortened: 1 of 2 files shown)


markbackman (approved):
LGTM

PR #16, opened 2025-03-06
[PCC-90] Accept custom request bodies in base image

cbrianhill:
This loosens the validation on request body so any JSON body is accepted.


vr000m (approved):
LGTM

PR #17, opened 2025-03-11
include pipecatcloud and make use of the new agent session arguments

aconchillo:
This now includes `pipecatcloud` package and makes use of the new arguments.


pipecat-base/requirements.txt
```diff
@@ -1,2 +1,3 @@
 fastapi[standard]~=0.115.0
 loguru~=0.7.3
+pipecat-cloud
```
markbackman, line 3: Typo:

```suggestion
pipecatcloud
```

markbackman (approved):
LGTM—just a tiny typo fix.

pipecat-base/requirements.txt
aconchillo: Ooops, done.

cbrianhill (approved):
This looks good.  It'll be a breaking change for folks that are building their bots from the `latest` tag, so we might want to bump the `VERSION` file in case anyone pinned to its current value.

markbackman (approved):
LGTM! 🚢 

Built the base image locally and ran the voice and twilio starters with success 🎉

PR #18, opened 2025-03-12
pipecat-base(app): use multiprocessing for Daily bots

aconchillo:
This is to avoid an issue related to `daily-python` regarding configuring virtual devices within the same process.

### DIFF
--- a/pipecat-base/app.py
+++ b/pipecat-base/app.py
@@ -1,23 +1,22 @@
 #!/usr/bin/env python
+
 import asyncio
-import multiprocessing
-
-from os import environ
-from typing import Annotated, Any
 import sys
+from multiprocessing import Process
+from os import environ
+from typing import Annotated
 
-from fastapi import FastAPI, Body, Header, WebSocket
+from fastapi import FastAPI, Header, WebSocket
 from fastapi.websockets import WebSocketState
-
 from loguru import logger
 
-from waiting_server import WaitingServer, Config
-
 from bot import bot
-
+from waiting_server import Config, WaitingServer
 
 app = FastAPI()
-server_config = Config(environ.get("SHUTDOWN_TIMEOUT", 7200), app, host="0.0.0.0", port=int(environ.get("PORT", 8080)))
+server_config = Config(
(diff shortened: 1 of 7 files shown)


pipecat-base/app.py
```diff
@@ -28,26 +27,57 @@
...
 
 @app.post("/bot")
-async def handle_bot_request(body: dict, x_daily_room_url: Annotated[str | None, Header()] = None, x_daily_room_token: Annotated[str | None, Header()] = None, x_daily_session_id: Annotated[str | None, Header()] = None, proxy_connection: Annotated[str | None, Header()] = None):
-    await run_default_bot(body, x_daily_room_url, x_daily_room_token, x_daily_session_id)
+async def handle_bot_request(
+    body: dict,
+    x_daily_room_url: Annotated[str | None, Header()] = None,
+    x_daily_room_token: Annotated[str | None, Header()] = None,
+    x_daily_session_id: Annotated[str | None, Header()] = None,
+    proxy_connection: Annotated[str | None, Header()] = None,
```
cbrianhill, line 39: This header can be removed, as it's not used anywhere.

pipecat-base/app.py
```diff
@@ -28,26 +27,57 @@
...
+
+    asyncio.run(default_bot_main(body, x_daily_room_url, x_daily_room_token, x_daily_session_id))
+
+
+def launch_default_bot(body, x_daily_room_url, x_daily_room_token, x_daily_session_id):
+    process = Process(
+        target=default_bot_process,
+        args=(body, x_daily_room_url, x_daily_room_token, x_daily_session_id),
+    )
+    process.start()
```
cbrianhill, line 67: I think we won't have issues running this way.  We will need to test in a deployed environment to be sure, but the multiprocessing module I used earlier was a `ProcessPoolExecutor`.  It looks like this boils down to `fork()` so I suspect we'll be fine.

cbrianhill (approved):
Looks good.  What's the formatter you used?  I'm thinking we should all use the same thing in order to avoid churn.

pipecat-base/app.py
aconchillo: Fixed.

PR #21, opened 2025-03-15
Base image HTTP server should wait for bot completion

cbrianhill:
This is a partial fix for missing billing info for Daily room bots.  We're adding a `Process.join()` and delaying HTTP response until that has completed.


aconchillo:
LGTM! Remove `print("waiting")` and `process.join()`. The while loop already makes sure we wait for the process to complete.

aconchillo (approved):

PR #24, opened 2025-03-26
pipecat-base: top-level bot module import and remove Process

aconchillo:
This PR removes the need to create a process to run agents that requires a Daily transport. This will need `daily-python` 0.16.1 (yet to be released).

However, the DailyTransport (which uses `daily-python`) still requires that only a single session is run simultaneously in the same process (therefore, in the same k8s pod), which I believe is guaranteed by Pipecat Clout. @cbrianhill to confirm.

### DIFF
--- a/pipecat-base/app.py
+++ b/pipecat-base/app.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python
 
-import asyncio
 import sys
-from multiprocessing import Process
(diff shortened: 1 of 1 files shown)


markbackman (approved):
LGTM!

aconchillo:
This is working as expected.

PR #28, opened 2025-03-27
PSTN/SIP image updates

markbackman:
This PR attempts to make the file more readable by organizing the dial-in and dial-out functionality into their own respective classes.

It also includes some typo and linting fixes to the README and bot.py file.

Lastly, I added a little bit of detail in places in the README to remove ambiguity.

I tested both implementations with this file.

### DIFF
--- a/pipecat-starters/pstn_sip/bot.py
+++ b/pipecat-starters/pstn_sip/bot.py
@@ -38,8 +38,161 @@ async def terminate_call(
     await result_callback("Goodbye")
 
 
+class DialInHandler:
+    """Handles all dial-in related functionality and event handling.
+
+    This class encapsulates the logic for incoming calls and handling
+    all dial-in related events from the Daily platform.
+    """
+
+    def __init__(self, transport, task, context_aggregator):
+        """Initialize the DialInHandler.
+
+        Args:
+            transport: The Daily transport instance
+            task: The PipelineTask instance
+            context_aggregator: The context aggregator for the LLM
+        """
+        self.transport = transport
+        self.task = task
+        self.context_aggregator = context_aggregator
+        self._register_handlers()
+
+    def _register_handlers(self):
+        """Register all event handlers related to dial-in functionality."""
+
+        @self.transport.event_handler("on_dialin_ready")
+        async def on_dialin_ready(transport, data):
+            """Handler for when the dial-in is ready (SIP addresses registered with the SIP network)."""
+            # For Twilio, Telnyx, etc. You need to update the state of the call
+            # and forward it to the sip_uri.
+            logger.debug(f"Dial-in ready: {data}")
+
+        @self.transport.event_handler("on_dialin_connected")
+        async def on_dialin_connected(transport, data):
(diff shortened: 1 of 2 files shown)


pipecat-starters/pstn_sip/bot.py
```diff
@@ -150,132 +294,34 @@ async def main(room_url: str, token: str, body: dict):
...
-        nonlocal dialout_attempt_count
-        dialout_attempt_count += 1
-
-        if dialout_attempt_count > max_attempts:
-            logger.error(f"Max dialout attempts ({max_attempts}) reached, giving up")
-            return
-
-        logger.debug(f"Dialout attempt {dialout_attempt_count}/{max_attempts}")
+    # Initialize handlers dict to keep references
+    handlers = {}
```
markbackman, line 305: This is to resolve a type warning by my IDE. Open to other ways to solve this. The handlers are only used for their references to the registered handlers.

pipecat-starters/pstn_sip/bot.py
```diff
@@ -128,15 +273,11 @@ async def main(room_url: str, token: str, body: dict):
     context = OpenAILLMContext(messages, tools)
     context_aggregator = llm.create_context_aggregator(context)
 
-    # RTVI events for Pipecat client UI
-    rtvi = RTVIProcessor(config=RTVIConfig(config=[]))
```
markbackman, line 132: RTVI isn't relevant since this is a phone bot. Removing it from all places.

pipecat-starters/pstn_sip/bot.py
```diff
@@ -66,11 +215,7 @@ async def main(room_url: str, token: str, body: dict):
             f"Dialin settings: To: {dialled_phonenum}, From: {caller_phonenum}, dialin_settings: {dialin_settings}"
         )
 
-    voicemail_detection = bool(body.get("voicemail_detection"))
```
markbackman, line 69: I found this to be confusing, so I'm recommending that it be removed.

vr000m: I was hoping @DominicStewart would add the voicemail detection, but we can do this later

vr000m:
> This PR attempts to make the file more readable by organizing the dial-in and dial-out functionality into their own respective classes.
> 
> It also includes some typo and linting fixes to the README and bot.py file.
> 
> Lastly, I added a little bit of detail in places in the README to remove ambiguity.
> 
> I tested both implementations with this file.

I kept the dial-in and dial-out all in one place, so that we could reuse it in the transfer scenarios. Now, depending on the transfer criterion the handlers would have to be initiated.

vr000m (approved):
LGTM

markbackman:
> I kept the dial-in and dial-out all in one place, so that we could reuse it in the transfer scenarios. Now, depending on the transfer criterion the handlers would have to be initiated.

This is meant to be a starter, not a complete example, so I think it makes sense to optimize for distinctly showing how to do things and not overcomplicating with features. We can build demos using the starter images to show things like transfers and voicemail detection.

PR #29, opened 2025-04-06
Dom/voicemail detection

DominicStewart:
I've added the voicemail detection logic to the example. I've based it off of the gemini example I made where we have one pipeline for voicemail detection and another for the human conversation after. This allows us, if we want, to have two LLMs for different parts of the process. 

The task was to look if we have voicemail_detection in the body and if so, use that logic. Because of how many if statements we'd probably have and repeated logic, I decided to move the pipeline creation to its own class too. Happy to take it out of a class and put the various logic pieces in the main function, but thought it might get a bit long.


vr000m:
@DominicStewart FYI, I found an error in the README.md, which I fixed and pushed to your branch.

vr000m (commented):
LGTM

PR #32, opened 2025-04-10
PCC-181: Add Krisp support in base image

sunahsuh:


mattshep (approved):
I assume this works even though the symlink target doesn't exist until later?

sunahsuh:
Yep – if the flag is false and you try to import krisp it'll just complain about the library not existing, but if it's set to true then it's importable and usable in your script

markbackman (approved):
Works great!

PR #33, opened 2025-04-11
Make the base Pipecat image's Python version configurable

mattmatters:
Noticed this was at 3.10 but wanted to run a later version of Python. This also enables you to run a matrix build job and push a bunch of different python versions of your base image.

Happy to contribute a PR for that too.


markbackman:
@mattmatters we just added official support for this:
https://github.com/daily-co/pipecat-cloud-images/pull/43

Thanks for the PR. You were way ahead of us!

PR #39, opened 2025-05-13
Update TransportParams to match the latest Pipecat

markbackman:
Complies with the latest changes from Pipecat 0.0.66:

```
- `TransportParams.vad_enabled` parameter is now deprecated, use
  `TransportParams.audio_in_enabled` and `TransportParams.vad_analyzer` instead.

- `TransportParams.vad_audio_passthrough` parameter is now deprecated, use
  `TransportParams.audio_in_passthrough` instead.
```

Note: `audio_in_passthrough` now defaults to True, so it can be removed from the params.


filipi87 (approved):
LGTM.

PR #41, opened 2025-05-14
[PCC-291] Use exec when running python to fix signal handling on termination

cbrianhill:
This pull request includes a small but significant change to the `CMD` instruction in the `pipecat-base/Dockerfile`. The change ensures that the Python process replaces the shell process by using `exec`, which improves process management in containerized environments.

* [`pipecat-base/Dockerfile`](diffhunk://#diff-211cad0b3b0f9515bfe841cf3d3bad56ff7bb6150a15c155b22654de1ef883a9L10-R10): Updated the `CMD` instruction to use `exec` when running `python app.py`, ensuring the Python process becomes the main process in the container.


mattshep (approved):
💡

kompfner (approved):
Nice catch! Love a one-word fix.

PR #59, opened 2025-08-22
Migrate pipecat-base to uv

markbackman:
I'm updating the pipecat-base image to be built with uv. We should see a number of benefits:
- Faster build times
- More reproducible builds, thanks to `uv.lock`
- Faster container startup due to bytecode compliation
- Better caching

(At least these are the benefits that uv writes about!)

Here are the docs:
https://docs.astral.sh/uv/guides/integration/docker/

Examples Dockerfile:
https://github.com/astral-sh/uv-docker-example/blob/main/Dockerfile

Details:
- This updates the base image to an equivalent uv base image: `python:3.12-slim` to `ghcr.io/astral-sh/uv:python3.12-trixie-slim` (`python3.12-trixie-slim` _should_ be the same Python base image as `python:3.12-slim`)
- Removed requirements.txt as pyproject.toml is used by uv
- Moved versioning into pyproject.toml
- Updated README with release steps


mattshep (approved):
I haven't tried it out yet but looks right to me, I definitely need to start using uv more!

markbackman:
I just built a few bots with this and all seems well. I'll do some more testing before releasing. Thanks for the chat earlier and the quick review ⚡

PR #66, opened 2025-09-12
Add body to WebsocketSessionArguments

markbackman:
DO NOT MERGE!

Wait for corresponding Pipecat Cloud PR to be released: https://github.com/daily-co/pipecat-cloud-sandbox/pull/265

---

Receives custom body data from PCC and injects it to bot() as `body` in the `WebsocketSessionArguments`.


cbrianhill (approved):
This looks good to me.  Only thing is that this requires the body query string to be encoded JSON.  I think that's a reasonable assumption, but we should make sure we call it out in our docs.

markbackman:
Ok! I've tested Twilio, Telnyx, and Plivo in production to confirm all is well. Merging 🚢

PR #67, opened 2025-09-17
Created an example showing how to use SmallWebRTC inside Pipecat Cloud.

filipi87:
Created an example showing how to use SmallWebRTC inside Pipecat Cloud.


pipecat-starters/smallwebrtc/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+# Only for local development with the local versions of Pipecat and Pipecat Cloud
+# First you need to build both Pipecat and Pipecat Cloud
+# python -m build
+COPY ./pipecatcloud-0.2.4.tar.gz pipecatcloud-0.2.4.tar.gz
+RUN pip install ./pipecatcloud-0.2.4.tar.gz
+COPY ./pipecat_ai-0.0.86.dev38.tar.gz pipecat_ai-0.0.86.dev38.tar.gz
+RUN pip install ./pipecat_ai-0.0.86.dev38.tar.gz
+COPY ./.env .env
+# --------- end of local development
```
filipi87, line 24: TODO: remove it. Only for local testing.

pipecat-starters/smallwebrtc/client/src/app.ts
```diff
@@ -0,0 +1,205 @@
...
+    await this.pcClient.initDevices()
+    this.connectBtn.disabled = true;
+    try {
+      this.updateStatus('Starting the bot');
+      const headers = new Headers();
+      headers.append("Authorization", `Bearer ${this.apiKey}`);
+      const startBotResult = await this.pcClient.startBot({
+          endpoint: this.startUrl,
+          headers: headers
+      })
```
filipi87, line 178: Need to add a comment that it is not supposed to be used like this in production. It is just for making local tests easier.

PR #68, opened 2025-09-19
Deprecate pipecat-starter images

markbackman:
We've outgrown these images. We have better examples now in the [pipecat-examples repo](https://github.com/pipecat-ai/pipecat-examples). All examples in pipecat-examples are being migrated to be Pipecat Cloud deployable. All demos referenced in the deprecation notice are already deployable to Pipecat Cloud.


cbrianhill (approved):
sounds good to me!

filipi87 (approved):
Makes sense.

PR #69, opened 2025-09-24
Adding support for WhatsApp in the base image.

filipi87:
- Adding support for WhatsApp in the base image.
- Adding support for the mock server to emulate how Pipecat Cloud should handle WhatsApp webhooks.

### DIFF
--- a/pipecat-base/app.py
+++ b/pipecat-base/app.py
@@ -24,9 +24,7 @@ from waiting_server import Config, WaitingServer
 
 try:
     from pipecatcloud.agent import SmallWebRTCSessionArguments
-
     SMALL_WEBRTC_AVAILABLE = True
-    logger.info("SmallWebRTCSessionArguments available.")
 except ImportError:
     SMALL_WEBRTC_AVAILABLE = False
     logger.warning("SmallWebRTCSessionArguments not available. Requires pipecatcloud>=0.2.5.")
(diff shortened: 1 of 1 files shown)


local-test/mock_pipecat_cloud.py
```diff
@@ -116,7 +149,107 @@ async def proxy_request(agent_name: str, session_id: str, path: str, request: Re
             )
 
 
+# ---------------- WhatsApp specific routes ----------------
+@app.get(
+    "/whatsapp",
+    summary="Verify WhatsApp webhook",
+    description="Handles WhatsApp webhook verification requests from Meta",
+)
+async def verify_webhook(request: Request):
```
filipi87, line 158: @cbrianhill, if I’m not mistaken, as discussed on slack, we need to implement an equivalent endpoint inside the API service (TypeScript), and in this case, the verification token will be the Pipecat public API key.

filipi87: PR adding support to the sandbox: https://github.com/daily-co/pipecat-cloud-sandbox/pull/274

cbrianhill (approved):
I think this looks good.  One thing I'd make sure to test before merging is that when the bot is killed, it waits for any ongoing call to complete.  I generally think it will be just fine, but since we're touching the lifecycle of the server, it's a good practice to double-check.

PR #70, opened 2025-10-03
Fix: bump pipecatcloud client to 0.2.5

markbackman:
Without a constraint to `pipecatcloud`, the `uv.lock` file was using the previous version, which was 0.2.3. But, app.py has:
```
from pipecatcloud.agent import (
    DailySessionArguments,
    PipecatSessionArguments,
    SessionArguments,
    SmallWebRTCSessionArguments,
    WebSocketSessionArguments,
)
```

which includes `SmallWebRTCSessionArguments` which wasn't released until 0.2.5. To fix this, we should bump the `pipecatcloud` dependency to 0.2.5.

I'm not version bumping since 0.1.5 is broken and this will fix it.

### DIFF
--- a/pipecat-base/pyproject.toml
+++ b/pipecat-base/pyproject.toml
@@ -6,12 +6,15 @@ requires-python = ">=3.10"
 dependencies = [
     "fastapi[standard]~=0.115.0",
     "loguru~=0.7.3",
-    "pipecatcloud",
+    "pipecatcloud>=0.2.5",
     "pip",
 ]
 
 [tool.uv]
 dev-dependencies = []
+
+[tool.hatch.build.targets.wheel]
+packages = ["."]
 
 [build-system]
 requires = ["hatchling"]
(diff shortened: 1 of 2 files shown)


pipecat-base/pyproject.toml
```diff
@@ -6,13 +6,16 @@ requires-python = ">=3.10"
...
     "loguru~=0.7.3",
-    "pipecatcloud",
+    "pipecatcloud>=0.2.5",
     "pip",
 ]
 
 [tool.uv]
 dev-dependencies = []
 
+[tool.hatch.build.targets.wheel]
```
markbackman, line 16: uv sync was failing. Hatchling expects either:
- A directory named pipecat_base (matching your package name), or
- Explicit configuration telling it which files to include

Since this isn't a python package, I think this makes sense. It allows uv sync to succeed. I could be wrong though!

aconchillo (approved):

filipi87 (approved):
Makes sense.

PR #72, opened 2025-10-04
Make app.py more flexible based on route type

markbackman:
- Lazy load related imports
- Reorganize /api/offer such that all imports and configuration variables are initialized within the route

This avoids errors/warnings appear in logs for routes that aren't used.

### DIFF
--- a/pipecat-base/app.py
+++ b/pipecat-base/app.py
@@ -11,23 +11,9 @@ from bot import bot
 from fastapi import BackgroundTasks, Header, HTTPException, Query, WebSocket
 from fastapi.websockets import WebSocketState
 from loguru import logger
-from pipecatcloud.agent import (
-    DailySessionArguments,
-    PipecatSessionArguments,
-    SessionArguments,
-    WebSocketSessionArguments,
-)
+from pipecatcloud.agent import SessionArguments
 from pipecatcloud_system import app
 from waiting_server import Config, WaitingServer
-
-try:
-    from pipecatcloud.agent import SmallWebRTCSessionArguments
-
-    SMALL_WEBRTC_AVAILABLE = True
-    logger.info("SmallWebRTCSessionArguments available.")
-except ImportError:
-    SMALL_WEBRTC_AVAILABLE = False
-    logger.warning("SmallWebRTCSessionArguments not available. Requires pipecatcloud>=0.2.5.")
 
 server_config = Config(
     environ.get("SHUTDOWN_TIMEOUT", 7200),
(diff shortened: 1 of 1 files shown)


pipecat-base/app.py
```diff
@@ -121,108 +125,121 @@ async def handle_websocket(
 
 
 # ------------------------------------------------------------
-# Optional: SmallWebRTC route only if pipecat is available
+# SmallWebRTC route
 # ------------------------------------------------------------
-if SMALL_WEBRTC_AVAILABLE:
+@app.post("/api/offer")
+async def offer(
+    request: dict,
```
markbackman, line 132: Changing to a `dict` so that we can avoid defining the `SmallWebRTCRequest` before the route.

filipi87:
Hi @markbackman, I believe with the change in this PR, we are creating a new handler each time we receive a new offer.

This means it won’t prevent multiple WebRTC connections from being created, and it also won’t work properly when the offer endpoint is called again for renegotiation, instead of reusing the existing connection, it will try to create a new one. So, I believe it is going to be an issue for the SmallWebRTC.

I guess our previously approach was right. The issue was that when Pipecat was available, but if an internal import problem occurs within Pipecat, it was raising a general `Exception` instead of an `ImportError`.

I believe we should implement a fix as simple as this one:
- https://github.com/daily-co/pipecat-cloud-images/pull/73

What do you think ?

markbackman:
> Hi @markbackman, I believe with the change in this PR, we are creating a new handler each time we receive a new offer.
> 
> This means it won’t prevent multiple WebRTC connections from being created, and it also won’t work properly when the offer endpoint is called again for renegotiation, instead of reusing the existing connection, it will try to create a new one. So, I believe it is going to be an issue for the SmallWebRTC.
> 
> I guess our previously approach was right. The issue was that when Pipecat was available, but if an internal import problem occurs within Pipecat, it was raising a general `Exception` instead of an `ImportError`.
> 
> I believe we should implement a fix as simple as this one:
> 
> * [Handling Pipecat-ai initialization failure by disabling WebRTC route. #73](https://github.com/daily-co/pipecat-cloud-images/pull/73)
> 
> What do you think ?

Agreed. Handling the Exception raised by the ModuleNotFoundError in pipecat.transports.smallwebrtc.connection is the root cause. I added this additional lazy loading because logs will have errors and warnings about SmallWebRTC, even when it's not in use. But, as you point out, that has a downside, which we likely want to avoid.

Closing in favor of #73.

PR #73, opened 2025-10-05
Handling Pipecat-ai initialization failure by disabling WebRTC route.

filipi87:
Handling Pipecat-ai initialization failure by disabling WebRTC route.


markbackman (approved):
LGTM

PR #93, opened 2026-04-16
Add Python 3.14 support, deprecate 3.10

markbackman:
We can drop 3.10 once it's EOL. Developers on `pipecat-ai` <1 can still use those images. I think this is harmless to carry forward for now, but let me know if you disagree.

Also, I'd like to make the default image 3.13, but I think we can do that separately.

## Summary
- Adds Python 3.14 to the build matrix, aligning with Pipecat 1.0.0's supported versions
- Deprecates Python 3.10 images (to be removed in a future release)
- Updates dependency bounds: `fastapi` upper bound to `<1`, `pipecatcloud` to `>=0.4.4`, `nltk` to `>=3.9.4`
- Bumps version to 0.1.18

## Test plan
- [ ] CI builds images for all five Python versions (3.10, 3.11, 3.12, 3.13, 3.14)
- [ ] Default `latest` tag still points to Python 3.12
- [ ] Existing Python 3.10–3.13 images continue to work

🤖 Generated with [Claude Code](https://claude.com/claude-code)


filipi87 (approved):
Makes sense. 🚀

PR #98, opened 2026-07-21
Structured log emission for the PCC log-collection lane (PCC-1024)

mattshep:
Plan item **L2** of the on-prem logging plan ([logging-plan.md](https://github.com/daily-co/pipecat-cloud-region-package/blob/main/docs/logging-plan.md)) — the emit side of the log-collection lane.

## What

When the platform sets `PCC_LOG_DIR` (the operator does, for regions with logging enabled), the runner writes structured JSONL to `$PCC_LOG_DIR/bot.jsonl` for an in-pod shipper to tail. **Unset → byte-identical behavior** — cloud and non-PCC users are unaffected.

Record schema (the shipper adds pod-constant fields like namespace/service/deployment): `@timestamp`, `stream` (`app`|`stdout`|`stderr`), `level` (app lane only), `session_id`, `line`.

## How

- **Framework lane:** one loguru file sink (rotation/retention/`enqueue` for free); `session_id` rides the existing `contextualize` threading.
- **Capture lane:** `pcc_structured_logs.install()` runs **before `import bot`** (customer code may print at import time) and dup2's pipes over stdout/stderr. Pump threads pass the original bytes through to the real streams — `kubectl logs` and any node collector see exactly what they saw before — and re-emit each line tagged `pcc_stream`, so raw `print()` and C-extension output join the lane.
- **Recursion safety:** no loguru console sink may write to a captured fd. `install()` swaps the default sink onto the saved real stderr; app.py's console sink uses `console_stream()` + `console_filter` (which also prevents captured lines rendering twice).
- **Session attribution with linger:** captured lines are stamped from a module slot set by `run_bot` (exact under enforced concurrency-1), cleared with a **250ms linger** — the pump emits asynchronously, so a bot's final pre-crash output keeps its attribution; a new session cancels the linger and takes over immediately. *The subprocess e2e test caught this race; the linger is the fix.*
- Overlong captured lines are emitted truncated (16KB cap) then discarded to the next newline; passthrough always carries the full original bytes.

## Tests

New pytest suite (repo previously had no test runner) + a lint-and-test CI workflow:
- line assembler (splits, partials, CRLF, UTF-8 replacement, truncate/discard)
- serializer schema (field presence per lane, placeholder-session omission, braces/unicode, failure→marker-record)
- disabled no-op (everything inert without `PCC_LOG_DIR`)
- **subprocess end-to-end**: real fd capture in a child process asserting kubectl-logs equivalence (passthrough, no double-render) and schema-valid records for framework/print/stderr/out-of-session lines

`ruff format --check`, `ruff check`, `pytest` (19 passed) all green; `docker-publish.yml` builds the image on PR (validates the Dockerfile COPY change).

Note: GitHub flags 3 pre-existing Dependabot vulns on the default branch — unrelated to this change.


cbrianhill (approved):

markbackman (approved):
LGTM

PR #99, opened 2026-07-22
Make the structured-log file sink crash-durable (PCC-1038)

mattshep:
## What

The PCC-1032 crash-tail e2e showed the runner's structured-log sink losing the final pre-crash records: with `enqueue=True`, loguru routes records through a queue thread that `os._exit` destroys before draining — the last ~30ms of a crashing bot's output (the lines the crash tail exists for) never reached disk. Dying words survived **1 of 3** crash runs.

## Fix

`enqueue=False`. loguru's file sink is line-buffered (`mode='a', buffering=1`), so each record is a single synchronous `write()` into the OS **page cache** — which survives `os._exit`/SIGKILL; Fluent Bit reads the page cache even after the process is gone. Cost: µs-scale file I/O on the emitting thread, negligible at bot logging volumes (the original "never block the event loop" comment traded away exactly the records that matter most).

## Verified

- **Regression test with teeth**: new subprocess test hard-exits (`os._exit(17)`) immediately after its last `logger.error` — deterministically **fails under `enqueue=True`** (verified by flipping it back: `KeyError: 'these are the dying words'`, the exact production symptom) and passes now. Suite 20/20.
- **Docker acceptance repro** (base built from this branch + the logging-e2e fixture, `{"crash": true}` × 3): framework-lane dying words **3/3** (was 0/3).

## Scope honesty

Captured stdout/stderr keeps one unavoidable async hop — the pump thread must read the pipe before the process dies — so a `print()` in the final microseconds before a hard exit remains best-effort (0/3 in the repro; nothing fd-level can read a pipe after the writer's process is gone). Documented in the module docstring. The lane that matters is guaranteed: the runner's real crash reporting (`run_bot`'s exception handler) speaks through the logger, and bots that want guaranteed dying words should too.

CHANGELOG 0.1.24 + lockfile refresh included (the #98 lesson).

### DIFF
--- a/pipecat-base/CHANGELOG.md
+++ b/pipecat-base/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to the **Pipecat Cloud Base Images** will be documented in t
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.25] - 2026-07-27
+
+### Fixed
+
+- Structured log file sink writes synchronously (was queued): the final log
+  lines before a hard process death (`os._exit`, SIGKILL) now reach disk and
+  ship with the crash tail. Captured stdout/stderr keep one async pump hop and
+  remain best-effort in the final microseconds — dying words said through the
+  logger are guaranteed. (PCC-1038)
+
 ## [0.1.24] - 2026-07-27
 
 ### Changed
--- a/pipecat-base/pyproject.toml
+++ b/pipecat-base/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "pipecat-base"
-version = "0.1.24"
+version = "0.1.25"
 description = "Pipecat Cloud Base Image"
 requires-python = ">=3.10"
 dependencies = [
--- a/pipecat-base/pcc_structured_logs.py
+++ b/pipecat-base/pcc_structured_logs.py
@@ -33,6 +33,14 @@ console sink may ever write to the *captured* descriptors — ``install()``
 removes loguru's default sink before capturing, and every console sink must
 use ``console_stream()`` (the saved real stderr) with ``console_filter``
 (which also keeps captured lines from being displayed twice).
+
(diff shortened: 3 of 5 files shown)


cbrianhill (approved):

pipecat-base/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "pipecat-base"
-version = "0.1.23"
+version = "0.1.24"
```
markbackman: You'll need to rebase and bump this to 0.1.25. I just merged a PR that bumped package numbers.
```suggestion
version = "0.1.25"
```

pipecat-base/CHANGELOG.md
```diff
@@ -5,6 +5,16 @@ All notable changes to the **Pipecat Cloud Base Images** will be documented in t
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.24] - 2026-07-22
```
markbackman: Same to bump version and update date to whenever this gets merged.

```suggestion
## [0.1.25] - 2026-07-27
```

mattshep:
Rebased onto main and bumped to **0.1.25** (`5bd4ef2`) — thanks @markbackman.

- `pyproject.toml`: 0.1.25, and `uv.lock` regenerated via `uv lock` so its self-referential `pipecat-base` version matches rather than being hand-edited.
- `CHANGELOG.md`: new `## [0.1.25] - 2026-07-27` section carrying the PCC-1038 entry, sitting above your 0.1.24 section (nltk bump + security), which is preserved as-is. Date is today, i.e. merge day.

Worth noting the rebase quietly absorbed the original version bump: the branch had bumped 0.1.23 → 0.1.24, and since main already landed 0.1.24 that hunk became a no-op, leaving pyproject at 0.1.24 with no conflict to flag it. Only the CHANGELOG actually conflicted. So the explicit 0.1.25 bump was needed in both files, not just the one that showed a conflict.

20 tests pass locally; all six CI checks green (test, find-images, base image builds 3.11–3.14).

markbackman (approved):

PR #101, opened 2026-07-25
Fix security vulnerabilities in nltk, pip, and pydantic-settings

markbackman:
## Summary

- Bumped `pipecat-base` to **0.1.24**.
- Updated dependencies to resolve Dependabot alerts: nltk 3.9.4→3.10.0 (GHSA-p4gq-832x-fm9v, high), pip 26.1.1→26.1.2 (GHSA-wf93-45jw-7689), pydantic-settings 2.12.0→2.14.2 (GHSA-4xgf-cpjx-pc3j). nltk is a direct dependency, so its floor moved from `>=3.9.4` to `>=3.10.0`.
- Refreshed `pipecat-base/uv.lock`.

## Notes for reviewers

- **nltk 3.10.0 is a behavioral change, not just a patch.** The advisory fix adds `nltk/pathsec.py`, which sandboxes all nltk file access to directories listed in `nltk.data.path` (plus `NLTK_DATA`, `~/nltk_data`, `/usr/share/nltk_data`, tempdir). Writes outside those roots now raise `PermissionError`.
- The image is unaffected: the Dockerfile downloads `punkt_tab` to `/usr/local/share/nltk_data`, which is a default `nltk.data.path` entry and therefore an allowed root. I verified this on 3.10.0 — download and unzip complete end-to-end into an allowed root, and the runtime `nltk.data.find` lookup reads from one too. Downloading to a *non*-listed directory fails with `PermissionError`, drops into an interactive retry prompt, and exits 1, so this was worth confirming before the build.
- Downstream bots that read or download nltk data from a custom directory do need to authorize it (`nltk.data.path.append(...)` or `NLTK_DATA`). Called out in the changelog.
- nltk 3.10.0 adds a new runtime dependency, `defusedxml` 0.7.1.
- The lockfile's own `pipecat-base` entry was stale at 0.1.23 while `pyproject.toml` said 0.1.24. The Dockerfile runs `uv sync --locked`, which fails on that mismatch, so the image build would have broken. Fixed in the lock refresh.
- Verification: 19 tests pass, `ruff check` clean, `uv lock --check` clean in both the root and `pipecat-base/`. Docker build not run locally (no daemon); CI's PR build covers it.

## Not covered here

Two Dependabot alerts remain against the **root** `uv.lock`. Both are dev tooling only — nothing that ships in the image:

- `pipecat-ai` 0.0.108 (2 high) — present only in the `python_full_version < '3.11'` resolution fork; 3.11+ already resolves to 1.6.0. Clearing it means raising the root `requires-python` to `>=3.11`, which lines up with the Python 3.10 deprecation noted in 0.1.18 but is a separate change.
- `pytest` (medium) — needs 9.0.3; the root pins `pytest~=8.4`.

## Release

Merging to `main` cuts the release — `docker-publish.yml` reads the version from `pipecat-base/pyproject.toml` and builds/pushes the `0.1.24` and `latest` tags. The changelog currently dates 0.1.24 as 2026-07-27; worth adjusting at merge time if that slips.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pipecat-base/CHANGELOG.md
+++ b/pipecat-base/CHANGELOG.md
@@ -5,6 +5,22 @@ All notable changes to the **Pipecat Cloud Base Images** will be documented in t
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.24] - 2026-07-27
+
+### Changed
+
+- Bumped minimum `nltk` from `>=3.9.4` to `>=3.10.0`. nltk 3.10.0 sandboxes
+  file access to directories listed in `nltk.data.path`. The base image's
+  pre-downloaded `punkt_tab` data is unaffected, but bots that read or
+  download nltk data from a custom directory must now authorize it via
+  `nltk.data.path.append(...)` or `NLTK_DATA`, or nltk raises
+  `PermissionError`.
+
+### Security
+
+- Update dependencies to resolve security vulnerabilities: nltk 3.9.4 to
+  3.10.0, pip 26.1.1 to 26.1.2, pydantic-settings 2.12.0 to 2.14.2.
+
 ## [0.1.23] - 2026-07-21
 
 ### Added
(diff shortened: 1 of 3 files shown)


filipi87 (approved):
LGTM!

cbrianhill (approved):

PR #102, opened 2026-07-31
feat(pipecat-base): enforce the platform's session budget in-process (PCC-1066)

mattshep:
The base-image half of [PCC-1066](https://linear.app/dailyco/issue/PCC-1066). Pairs with **[sandbox#719](https://github.com/daily-co/pipecat-cloud-sandbox/pull/719)**, which is standalone and can merge in either order.

## The problem

Pipecat Cloud caps sessions at `maxSessionDuration`, but from outside the container all it can do is close its own `/bot` request. On HTTP transports the media runs directly between this process and the transport provider, so that close is invisible to `bot()` and the session carries on. The pod then looks idle and takes another session into the same Python process.

Observed in production (Inderes, 2026-07-29): session `b29a7ff7` hit the 2h cap at 13:06:32, kept serving its broadcast until 13:49, and a second session was placed onto the same process at 13:21. Both died in an OOM at 13:51. The base image logs a matched `Starting bot session` / `Stopping bot session` pair per session — those two sessions each have a `Starting` and no `Stopping`, ever.

sandbox#719 stops the pod being handed out while a bot is live. This PR makes the cap actually stop the bot, which is only possible here.

## What this does

The platform sends `X-PCC-Max-Session-Seconds` on `/bot`; we enforce it around `await bot(args)`.

**No default.** Only the platform knows the configured value, and the limit can be set as high as 4 hours, so a hardcoded default would cut sessions short for anyone above it. An absent header means nothing is enforced, logged once so the inert case is diagnosable from the customer's own logs.

**The budget is stashed in `GLOBALS`, not passed down.** This looks like indirection but is load-bearing: SmallWebRTC calls `run_bot` **twice**. `/bot` only waits for the WebRTC connection and returns (`app.py:138-146`); the real `bot()` runs from the `/api/offer` background task (`app.py:349-355`), which never sees these headers. Passing the budget down from the handler would leave that path unbounded — and bounding the `/bot` wait *instead of* the bot would return the request while the bot was still running, i.e. reintroduce the exact bug this fixes. Same reason `pipecat_session_body` is already stashed.

**Cancellation goes through a small `_SessionRun` handle** rather than a bare `task.cancel()`, so the reason survives and cancellation we asked for is distinguishable from this process being torn down — the latter must still propagate. The budget is its only caller today; a platform "stop this session" request would reuse it without further plumbing. Not building that endpoint here: it would be an unauthenticated control surface on :8080 for a defect nobody has reported.

## Two behaviours pinned deliberately

Both were established by testing against real Pipecat before writing this, and both are easy to get wrong:

1. **A Pipecat bot absorbs the cancellation.** `WorkerRunner.run()` catches `CancelledError` at its `_shutdown_event.wait()` and then runs its *graceful* path — `CancelFrame` through the processors, `cleanup()` on each — in ~19 ms. So nothing is raised, and any "did the budget fire?" logic based on catching an exception silently misses the common case. That is why the reason is checked after the `await` as well as in the `except`.

2. **`run_bot` must return normally either way.** The HTTP response is what tells the platform the pod is free again. If the cancellation escaped instead, the platform would see a dropped connection and classify the session `BOT_DISCONNECTED` (PCC-892) — we would have built a machine that reports every capped session as a bot crash. Verified end-to-end against a real uvicorn server driven with the sidecar's client settings: clean `200` for both the absorbing and propagating bot shapes.

A bot that swallows cancellation and keeps running cannot be stopped by anything in-process — not even an outer `wait_for`. That case is covered by sandbox#719 keeping the pod marked occupied, which is why the two halves are designed together.

## Behaviour change

**Sessions that currently overrun their configured limit will now stop at it.** Scoped:

| Situation | Today | After |
|---|---|---|
| WebSocket transport | ends at the cap | unchanged |
| HTTP, platform-created Daily room | overruns ~5 min until room expiry (`cap + 300`) | ends at the cap |
| HTTP, customer-supplied room or transport | **overruns unboundedly** | ends at the cap |

Only the third row is a real change. Inderes are in it — their ~3h broadcasts against a 2h cap only worked because of the overrun. Workaround is to raise `maxSessionDuration` (max 4h), called out in the changelog.

Docs are corrected in a companion PR: the current wording says the bot "receives a connection error the next time it tries to send or receive data", which has never been true for HTTP transport and is exactly why the session didn't stop.

## Release ordering

**Default the new queue-sidecar before publishing this image.** Picking up a new base image requires a rebuild + redeploy, which stamps the then-current sidecar onto the new revision — so `old sidecar + new base image` is unreachable *only if* the sidecar default has already moved. Getting it wrong is not dangerous (this half is inert without the header), but a customer who rebuilds for the fix would see no change and no signal explaining why.

## Testing

`40 passed`, ruff format + lint clean. New `test_session_budget.py` (17 tests) covers both bot shapes, the within-budget case, absent/malformed/non-positive headers, the teardown-must-propagate rule, the SmallWebRTC stash, and the cancellation primitive.

Mutation-tested: forcing `budget = None` fails both enforcement tests. Worth noting the first pass of that mutation only failed *one* of them — an absorbing bot makes the test's own outer `wait_for` return normally rather than raise, so the test passed for the wrong reason. Both now assert on elapsed time.

### DIFF
--- a/pipecat-base/app.py
+++ b/pipecat-base/app.py
@@ -114,10 +114,10 @@ image_version = environ.get("IMAGE_VERSION", "unknown")
 
 # Header carrying the session budget Pipecat Cloud has configured for this
 # service (`maxSessionDuration`). Enforcing it here is the only way the cap can
-# actually stop the bot: the platform can close its own request to us, but on
-# HTTP transports the media runs directly between this process and the transport
-# provider, so closing that request is invisible to bot() and the session simply
-# carries on.
+# actually stop the bot: the platform can close its own request to us, but when
+# a session is started by an HTTP request the bot goes on to connect its own
+# media transport (a Daily room, say), so closing that request is invisible to
+# bot() and the session simply carries on.
 #
 # There is deliberately no default. Only the platform knows the configured
 # value, so an absent header means "not supplied" and nothing is enforced.
@@ -183,6 +183,23 @@ def _parse_budget(raw: Optional[str]) -> Optional[float]:
     return budget
 
 
+def _cancellation_requested_on_self() -> bool:
+    """Has something cancelled the task running this coroutine?
+
+    Distinguishes "we cancelled the bot" from "someone cancelled us", which
+    matters while the bot is unwinding after the budget fired.
+
+    ``Task.cancelling()`` is 3.11+. Base images are built for 3.10 too (the
+    default is 3.12), so on 3.10 we cannot tell the two apart and fall back to
+    treating the cancellation as ours. The cost is confined to a teardown that
+    lands inside the bot's unwind window — a few tens of milliseconds — on a pod
+    that is going away regardless.
+    """
+    task = asyncio.current_task()
+    cancelling = getattr(task, "cancelling", None)
+    return bool(cancelling and cancelling())
+
+
 async def _run_bot_with_budget(args: SessionArguments) -> None:
     """Run bot(args), stopping it if it outlives the platform's session budget.
 
@@ -213,7 +230,13 @@ async def _run_bot_with_budget(args: SessionArguments) -> None:
     except asyncio.CancelledError:
         # Only swallow the cancellation we asked for. Anything else means this
         # process is going away, and the caller needs to see that.
-        if run.cancel_reason is None:
+        #
+        # cancel_reason alone is not enough to tell them apart: it stays set for
+        # the rest of the session, so a teardown arriving while the bot is still
+        # unwinding would look like our own deadline. We cancel the *bot* task
+        # and never this one, so a pending cancellation against this task can
+        # only have come from outside.
+        if run.cancel_reason is None or _cancellation_requested_on_self():
             raise
     finally:
         if deadline_task is not None:
--- a/pipecat-base/app.py
+++ b/pipecat-base/app.py
@@ -16,7 +16,7 @@ import logging
 import sys
 from contextlib import asynccontextmanager
 from os import environ
-from typing import Annotated, Callable, List, Optional, Union
+from typing import Annotated, Callable, Dict, List, Optional, Union
 
 import aiohttp
 import bot as bot_module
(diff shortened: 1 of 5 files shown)


pipecat-base/app.py
```diff
@@ -112,6 +112,122 @@ def filter(self, record: logging.LogRecord) -> bool:
...
+
+        deadline_task = asyncio.create_task(_deadline())
+
+    try:
+        await run.task
+    except asyncio.CancelledError:
+        # Only swallow the cancellation we asked for. Anything else means this
+        # process is going away, and the caller needs to see that.
+        if run.cancel_reason is None:
+            raise
```
cbrianhill, line 240: Edge case, non-blocking, approving regardless.

Once the deadline fires, `cancel_reason` stays set for the rest of the session, so this arm swallows *any* `CancelledError` arriving afterwards, including process teardown. The invariant is enforced only up until the cap fires.

Repro (absorbing bot, 0.05s budget, 0.5s unwind, `task.cancel()` at 0.15s): the outer cancellation is swallowed and the coroutine returns normally.

`test_outer_cancellation_still_propagates` uses `budget = 30`, so `cancel_reason` is never set and the test only covers the pre-deadline case. That is why this is invisible today.

The discriminator is that the budget cancels the inner task and never the outer one, so `Task.cancelling()` separates them cleanly:

```python
except asyncio.CancelledError:
    outer = asyncio.current_task()
    if run.cancel_reason is None or (outer is not None and outer.cancelling()):
        raise
```

Verified both paths with that patch: budget-fires-no-teardown returns normally, budget-fires-then-teardown propagates.

Impact is genuinely small. The window is ~19ms for a normal Pipecat unwind, and the pod is terminating anyway, so the visible effect is a 200 instead of a propagated cancel. Flagging it mainly because it is one of the two behaviours the PR pins deliberately, and the test does not reach the case where it fails.

One caveat if you do take it: `Task.cancelling()` is 3.11+. The Dockerfile pins `PYTHON_VERSION=3.12`, but `pyproject.toml` says `requires-python = ">=3.10"`.

cbrianhill (approved):
Approving. The design is sound and the cross-PR contract with sandbox#719 holds up: the header is set after the customer-header merge so a caller cannot raise its own cap, it is a duration rather than an absolute deadline so there is no clock-skew coupling between the containers, and the sidecar's accounting deadline fires strictly before the in-process one, so the ordering its comments assume is actually guaranteed.

Verified locally: cloned the branch and ran `pytest tests/`, 40 passed as claimed.

One edge case inline, non-blocking.

Worth calling out: checking `cancel_reason` after the `await` as well as in the `except` is the non-obvious detail the whole change turns on, and the comment explains why rather than just what. The mutation-testing note about the first pass failing only one of the two tests is exactly the failure mode that would otherwise have shipped a green-but-meaningless test.

mattshep:
Good catch — reproduced it before changing anything: absorbing bot, 0.05s budget, 0.5s unwind, `task.cancel()` at 0.15s, and the outer cancellation is swallowed exactly as you describe.

You're right about the shape of it too: `cancel_reason` stays set for the rest of the session, so the invariant held up to the cap and not after it, and `test_outer_cancellation_still_propagates` used a 30s budget so it never reached the case. That's the second time on this ticket a test has passed for the wrong reason.

Taken your fix, with one change for the version caveat you flagged. `versions.yaml` does still build 3.10, so `Task.cancelling()` is guarded:

```python
def _cancellation_requested_on_self() -> bool:
    task = asyncio.current_task()
    cancelling = getattr(task, "cancelling", None)
    return bool(cancelling and cancelling())
```

On 3.10 that degrades to the previous behaviour, which I've documented at the call site — the exposure is a teardown landing inside the unwind window on a pod that is going away anyway. Bumping `requires-python` to 3.11 would fix it properly but that's a product decision about dropping a base image, not one for this PR.

Added `test_teardown_during_the_unwind_still_propagates`, which fails against the old discriminator and passes with the new one. Verified both paths directly as well: budget-fires-no-teardown returns normally, budget-fires-then-teardown propagates.

Also picked up two wording points from your review of the docs PR, since they apply to the code comments here: "HTTP transport" conflated activation with media (the HTTP request only *starts* the session — the bot then connects its own transport, which is exactly why closing that request told it nothing), and `cleanup()` running is conditional on the bot using `PipelineRunner`, so the changelog no longer promises it outright.

41 tests pass, ruff clean.

PR #104, opened 2026-09-09
Give PID 1 to tini and exit on a SIGTERM during startup (PCC-1004)

mattshep:
The base-image half of [PCC-1004](https://linear.app/dailyco/issue/PCC-1004). The queue-sidecar got this in [pipecat-cloud-sandbox#805](https://github.com/daily-co/pipecat-cloud-sandbox/pull/805); the two containers share a pod and a grace period, so with the sidecar alone the pod still lingered on the user container.

## Why

The kernel discards a signal to PID 1 that has no handler. The CMD execs Python into PID 1, and uvicorn only installs its SIGTERM handler once every module-level import in `app.py` has finished — the customer's `bot.py` included, whose import time is unbounded (the quickstart loads Smart Turn and Silero there). A SIGTERM in that window was dropped, and the container went on to start and serve on a pod the kubelet had already told to stop, until the grace-period SIGKILL up to 7500s later. That window isn't unlucky: a pod deleted during its image pull gets the kill dispatched the moment its containers start.

## What changed

Two layers, because customers build `FROM` this image and may replace the `ENTRYPOINT`:

- **tini is PID 1** (`ENTRYPOINT ["/usr/bin/tini", "-e", "143", "--"]`, from Debian so buildx resolves the arch). Python is its child and dies on the default disposition. tini also reaps children orphaned by `pre-app.sh` — the Datadog guide backgrounds an agent there — which Python as PID 1 never did.
- **`pcc_early_sigterm`** is the first statement of `app.py`: a handler that exits at once during startup (nothing to drain), which holds even when a derived image puts Python back at PID 1.

## Details worth a reviewer's attention

**No `-g`.** I checked what it would reach: Datadog's init scripts use `start-stop-daemon --background`, which puts the agent in its own process group, so `-g` never touches it. Where it *does* reach — a naive `cmd &` in `pre-app.sh` — it would kill that helper at the start of a drain that can last the whole session (`SHUTDOWN_TIMEOUT` defaults to 7200s). Signals go to Python only, as in #805.

**`-e 143`, and the handler absorbing uvicorn's re-raise — these are the same concern.** uvicorn re-raises the captured SIGTERM after its drain so the process "dies from it". As PID 1 that re-raise was silently discarded, so `run()` returned and the interpreter exited normally — asyncio teardown, atexit hooks (where OpenTelemetry, Sentry and ddtrace flush), exit 0. I traced every consumer of the exit code: the operator's `container_health.go`, the API's headline mapping, the dashboard card and CLI `agent status`, and the terminal pod phase that k8s 1.34 stamps on deletion and `unhealthy_pods.yaml` selects with no namespace filter. All of them have only ever seen `Completed` / `Succeeded` on a stop. So:

- the early handler stays installed and, once `server.should_exit` is set, **absorbs** the re-raise, reproducing the PID 1 exit path exactly. My first cut reset it to `SIG_DFL`; under tini the re-raise then killed the process on the spot and skipped every atexit hook — I measured a registered hook never running on graceful stops. There's a unit test that mimics the re-raise and asserts it's absorbed;
- `-e 143` covers what the handler can't: a signal in the interpreter's own startup, or a `CMD` that isn't `app.py`.

**`defer_to_server` leaves the handler alone.** A bot module (or an SDK it imports) may install its own SIGTERM handler at import time; uvicorn saves and restores whatever it finds, so that has to survive untouched. Tested.

**CI gates the push, not just the PR.** Every customer bot is `FROM` this image, so `build-base` now `needs` the signal-check job on push as well. Single-arch and loaded locally because a multi-platform build can't be loaded into the daemon; the behaviour is arch-independent and the checks also pass on an emulated amd64 build.

## Verification

`container_signal_check.py` against this image and against the published `:latest`:

```
                                         this image    dailyco/pipecat-base:latest
python is not pid 1                      ok            FAIL  (pid 1)
unhandled SIGTERM → exit 0               ok            FAIL  (survived)
handled SIGTERM reaches python           ok            ok
early SIGTERM during bot import, tini    ok            FAIL  (survived)
early SIGTERM, tini bypassed             ok            FAIL  (survived)
graceful stop drains through uvicorn     ok            ok
```

Then the real `pipecat-quickstart` (Daily transport, models loaded at import), built `FROM` this image and `FROM 0.1.27`, with a real room and a real remote peer:

- SIGTERM during model load: exits 0 at once here, with or without tini; **0.1.27 survives it and goes on to start the server**.
- SIGTERM mid-call: drains until the peer leaves, `/bot` answers 200, exit 0 — identical log tails on both bases apart from the PID.

46 unit tests (5 new), ruff clean.

## Noticed while testing, not addressed here

- A session that outlives `SHUTDOWN_TIMEOUT` hangs after the drain deadline: `asyncio.run`'s teardown cancels pipecat's internal tasks together with the bot task, so the unwind deadlocks until the grace SIGKILL. Pre-existing (PCC-989 made the wait loop give up; the teardown after it still hangs), reachable only when a customer lowers the documented knob below their session length. Unchanged by this PR.
- `PipelineRunner(handle_sigterm=True)` (default `False`) registers a loop handler that's never removed and never sets uvicorn's `should_exit`, so SIGTERM ends the session but the server keeps running. Pre-existing, unchanged.

## Reach

This reaches a customer only when they rebuild `FROM` the new base. The Datadog guide pins `dailyco/pipecat-base:0.0.7`.


pipecat-base/Dockerfile
```diff
@@ -30,13 +30,37 @@ RUN VENV_SITE_PACKAGES=$(/app/.venv/bin/python -c "import site; print(site.getsi
...
+# child on the default disposition, exit 143. Remapping that to 0 keeps what
+# every consumer of the exit code has always seen on a stop: Completed, and a
+# Succeeded pod, never Error / Failed. (After startup app.py itself absorbs
+# uvicorn's post-drain re-raise, so a normal stop exits 0 as it always did.)
+#
+# No -g: signals go to Python only. The shutdown drain can last a whole
+# session, and a helper backgrounded from pre-app.sh must not be killed at the
+# start of it; properly daemonised services are outside the process group
+# regardless.
+ENTRYPOINT ["/usr/bin/tini", "-e", "143", "--"]
```
cbrianhill: **What this PR does**

A Pipecat Cloud bot container that got told to stop while it was still starting up ignored the stop, went on to start serving, and stayed up until Kubernetes SIGKILLed it at the grace deadline, which for these pods can be up to 7500 seconds. This PR makes it exit instead.

**Why.** The kernel does not apply default signal dispositions to PID 1: a signal with no registered handler is discarded rather than killing the process. The image's `CMD` execs Python straight into PID 1, and uvicorn only installs its SIGTERM handler inside `Server.serve()`, which is after every module-level import in `app.py` has finished. That includes the customer's `bot.py`, whose import time is unbounded (the quickstart loads Smart Turn and Silero at import). Any SIGTERM landing in that window was dropped.

That window is not a rare unlucky race. A pod deleted while its image is still pulling has the kill dispatched the moment its containers start, so the signal arrives at container start and beats the interpreter essentially every time. PCC-1004 has the traces: `Started container` and `Killing :: Stopping container` at the same second, first log line a second later, then zero shutdown lines for the pod's whole 7500s life.

This is the customer-image half of PCC-1004. The queue sidecar got the same fix in pipecat-cloud-sandbox#805; the two containers share a pod and a grace period, so fixing only the sidecar still left the pod lingering on the user container.

**How.** Two layers, because customers build `FROM` this image and some replace the `ENTRYPOINT`:

- `Dockerfile` installs Debian's `tini` and makes it the entrypoint: `ENTRYPOINT ["/usr/bin/tini", "-e", "143", "--"]`. Python becomes tini's child, so it takes the normal default disposition and dies. `-e 143` remaps death-by-SIGTERM to exit 0 so the pod still reads `Completed` / `Succeeded`, which is what every consumer of the exit code has always seen on a stop. No `-g`, so signals go to Python only and a helper backgrounded from `pre-app.sh` is not killed at the start of a drain that can last a whole session.
- `pcc_early_sigterm.py` is a new module imported and installed as the very first statement of `app.py`, before anything else. Its handler exits at once during startup, which covers the same window when a derived image puts Python back at PID 1.

The subtle part is the exit path. uvicorn re-raises the SIGTERM it captured after its drain, so the process "dies from it". As PID 1 that re-raise was discarded, so `run()` returned and the interpreter exited normally through asyncio teardown and atexit hooks. Under tini it would kill the process on the spot and skip all of that. So the early handler stays installed and, once `server.should_exit` is true, absorbs the re-raise, reproducing the old exit path exactly.

**Scope and blast radius.** Seven files, about 400 lines, most of it tests. Two of them are runtime: the Dockerfile entrypoint and 13 lines in `app.py`. The rest is a new 80-line module, a 251-line container-level check, 160 lines of unit tests, and a CI job that now gates `build-base` on those container checks for `push` as well as `pull_request`.

It changes the entrypoint of the image every customer bot is built `FROM`, but it only reaches a customer when they rebuild. The Datadog guide, for instance, pins `dailyco/pipecat-base:0.0.7`.

pipecat-base/CHANGELOG.md
```diff
@@ -5,6 +5,23 @@ All notable changes to the **Pipecat Cloud Base Images** will be documented in t
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
```
cbrianhill: **1. Version not bumped, so merging republishes `0.1.27` with different runtime behavior** - `suggestion`

The changelog entry lands under `[Unreleased]`, but `pipecat-base/pyproject.toml` still says `0.1.27`, and the publish workflow reads the version from exactly there:

```yaml
# .github/workflows/docker-publish.yml:20-33
- id: get-versions
  run: |
    BASE_VERSION=$(yq '.project.version' pipecat-base/pyproject.toml)
    ...
    echo "base-version=$BASE_VERSION" >> $GITHUB_OUTPUT
```

```yaml
# .github/workflows/docker-publish.yml, build-base -> Extract Docker metadata
tags: |
  type=raw,value=latest-py${{ matrix.python-version }}
  type=raw,value=${{ needs.find-images.outputs.base-version }}-py${{ matrix.python-version }}
  type=raw,value=latest,enable=${{ matrix.python-version == needs.find-images.outputs.default-python }}
  type=raw,value=${{ needs.find-images.outputs.base-version }},enable=${{ matrix.python-version == needs.find-images.outputs.default-python }}
```

So merging to `main` pushes `dailyco/pipecat-base:0.1.27`, `:0.1.27-py3.12` and the other per-Python variants over the release dated 2026-07-31, which is still sitting right below the new section in this same file:

```markdown
# pipecat-base/CHANGELOG.md:8
## [Unreleased]

### Fixed

- A stop signal that arrives while the agent is still starting up ...

## [0.1.27] - 2026-07-31
```

The awkward part is that `0.1.27` is this PR's own control. The description's verification table reads "SIGTERM during model load: exits 0 at once here, with or without tini; **0.1.27 survives it and goes on to start the server**." After the merge that sentence is no longer true of the tag it names.

There is a second consumer beyond humans reading tags. The image stamps its version as an OCI label:

```dockerfile
# pipecat-base/Dockerfile:66-68
ARG VERSION
ENV IMAGE_VERSION=$VERSION
LABEL daily.co/pipecat-cloud-base-version=$VERSION
```

and the operator reads that label back to decide how to configure the pod:

```go
// pipecat-cloud-operator/internal/imageprocessor/processor.go:318-320
// BaseImageVersionLabel is the OCI label used to identify pipecat cloud base image versions.
const BaseImageVersionLabel = "daily.co/pipecat-cloud-base-version"
```

```go
// pipecat-cloud-operator/internal/controller/pod_spec_builder.go:195-198
func buildContainerProbes(baseImageVersion string, probePort int) (*corev1.Probe, *corev1.Probe) {
	// Use HTTP probes for newer base images with /readyz and /livez endpoints.
	if supportsHTTPProbes(baseImageVersion) {
```

Nothing gates on tini today, so no operator misbehavior follows from this merge. The point is narrower: that label is the platform's only handle on which base-image behavior a customer image carries, and the whole mechanism assumes version determines behavior. Republishing makes `0.1.27` name two different runtimes, one of which exits on an early SIGTERM and one of which does not.

Honest caveat, since it cuts against me: this repo has republished the tag before. `7f6f875` (2026-08-11) switched `build-base` to multi-arch after 0.1.27 shipped, so the published `0.1.27` index already differs from what was pushed on release day. That one added a platform rather than changing what the container does on a signal, which is why I think this one is different in kind.

Suggested fix, in this PR:

```toml
# pipecat-base/pyproject.toml:3
version = "0.1.28"
```

```markdown
# pipecat-base/CHANGELOG.md:8
## [0.1.28] - 2026-09-09
```

If the intent is instead to batch this with other work and bump at release time, that is a fine answer too, but then it is worth saying so on the PR, because as it stands the merge publishes.

pipecat-base/pcc_early_sigterm.py
```diff
@@ -0,0 +1,80 @@
...
+        # uvicorn's post-drain re-raise. Absorb it, as PID 1 did, so run()
+        # returns and the interpreter exits normally.
+        return
+    try:
+        stream = _console()
+        stream.write(_MESSAGE)
+        stream.flush()
+    except Exception:
+        pass
+    os._exit(0)
```
cbrianhill, line 119: **2. The early-exit path leaves no record in the structured log lane, and drops captured import output** - `suggestion`

When a pod exits through this handler, the only evidence is one line on the container's stderr. The structured lane, which is the one the platform actually ships and queries, gets nothing.

```python
# pipecat-base/pcc_early_sigterm.py:44-65
def _console():
    # Once pcc_structured_logs has captured the process's stderr, fd 2 is its
    # pipe and a line written there would die with us in the pump thread. Its
    # saved console stream is the real stderr; fall back to sys.stderr when the
    # capture is not installed.
    logs = sys.modules.get("pcc_structured_logs")
    stream = logs.console_stream() if logs is not None else None
    return stream or sys.stderr


def _on_sigterm(signum, frame):
    if _server_handled is not None and _server_handled():
        # uvicorn's post-drain re-raise. Absorb it, as PID 1 did, so run()
        # returns and the interpreter exits normally.
        return
    try:
        stream = _console()
        stream.write(_MESSAGE)
        stream.flush()
    except Exception:
        pass
    os._exit(0)
```

Going through `console_stream()` rather than `sys.stderr` is correct and I would not change it. After `pcc_structured_logs.install()` has run, fd 2 is a pipe read by a pump thread, and a line written there dies with the process. `console_stream()` is a dup of the real stderr taken before the capture, so the line reaches `kubectl logs` and any node collector.

What it does not reach is `bot.jsonl`:

```python
# pipecat-base/pcc_structured_logs.py (module docstring)
# Two lanes feed one JSONL file (``$PCC_LOG_DIR/bot.jsonl``), which an in-pod
# shipper tails:
#
# - **framework lane** - a loguru file sink serializes every ``logger.*`` call.
# - **capture lane** - ``install()`` dup2's a pipe over the process's stdout and
#   stderr file descriptors ...
```

The handler writes to neither lane. It bypasses loguru entirely, and it writes to the saved fd rather than the captured one, so the capture lane does not see it either. When `PCC_LOG_DIR` is set, `bot.jsonl` has no record that the process exited during startup.

That matters because the structured lane is where this class of bug gets diagnosed. PCC-1004's own evidence section is counts out of regional OpenSearch: "Pods failing `Cannot connect to host 10.100.0.1:443`: 9,006 (462,121 log lines)". The next time a pod exits during startup and someone asks why, the query that would answer it comes back empty.

There is a second, smaller loss on the same path. `os._exit(0)` skips the pump threads, so anything the bot module printed during its import that is still in the capture pipe is discarded. `pcc_structured_logs` says as much about itself:

```python
# pipecat-base/pcc_structured_logs.py (module docstring)
# Capture-lane lines have one unavoidable async hop -
# the pump thread must read the pipe before the process dies - so a ``print()``
# in the final microseconds before a hard exit is best-effort.
```

Before this PR the process kept running, so those lines eventually drained. Now the import trace is truncated at exactly the moment you would want it. This is a small loss and the pod is going away regardless, but it is a change, and the changelog does not mention it.

**The obvious fix is not safe, which is worth stating.** Calling `logger.warning(...)` here would hit the file sink, which is deliberately synchronous for precisely this reason (`enqueue=False`, so each record is one `write()` into the page cache and survives `os._exit`). But loguru takes a per-handler lock that is not reentrant. If SIGTERM lands while the main thread is already inside a `logger.*` call, the handler runs on that same thread, tries to take the lock it already holds, and hangs. A deadlocked container that waits out the grace period is a worse outcome than a missing log line, and it is the exact failure this PR exists to remove.

A lock-free option that keeps the current guarantees: open the log file once in `install()` and append a pre-formatted record from the handler with a raw `os.write`, which is async-signal-safe and takes no Python-level lock.

```python
# in install()
_jsonl_fd = None
log_dir = os.environ.get("PCC_LOG_DIR")
if log_dir:
    try:
        _jsonl_fd = os.open(os.path.join(log_dir, "bot.jsonl"),
                            os.O_WRONLY | os.O_CREAT | os.O_APPEND, 0o644)
    except OSError:
        _jsonl_fd = None

# in _on_sigterm(), before os._exit(0)
if _jsonl_fd is not None:
    try:
        os.write(_jsonl_fd, _JSONL_RECORD)   # bytes, built once at import time
    except OSError:
        pass
```

where `_JSONL_RECORD` matches the schema `_serialize` emits, minus the timestamp if building one in the handler is not worth it:

```python
_JSONL_RECORD = (
    b'{"stream":"app","level":"WARNING",'
    b'"line":"SIGTERM received before the server was up; nothing to drain, exiting"}\n'
)
```

The other reasonable answer is to accept the gap and say so in the module docstring, so the next person querying for a startup exit knows to look at the raw container log instead. Either is fine; I would just rather it be a decision than an omission.

pipecat-base/pcc_early_sigterm.py
```diff
@@ -0,0 +1,80 @@
...
+    """Exit immediately on SIGTERM. Call before any other import in app.py."""
+    signal.signal(signal.SIGTERM, _on_sigterm)
+
+
+def defer_to_server(server_handled):
+    """Call right before serving. ``server_handled()`` must return True once the
+    server has handled a SIGTERM of its own (uvicorn: ``server.should_exit``);
+    such a signal is then absorbed rather than acted on. The handler itself is
+    left alone — a bot module may have installed its own at import time, and
+    uvicorn saves and restores whatever it finds."""
```
cbrianhill, line 132: **3. A bot that installs its own SIGTERM handler at import time re-opens the window, and the changelog does not say so** - `question`

```python
# pipecat-base/pcc_early_sigterm.py:73-80
def defer_to_server(server_handled):
    """Call right before serving. ``server_handled()`` must return True once the
    server has handled a SIGTERM of its own (uvicorn: ``server.should_exit``);
    such a signal is then absorbed rather than acted on. The handler itself is
    left alone - a bot module may have installed its own at import time, and
    uvicorn saves and restores whatever it finds."""
    global _server_handled
    _server_handled = server_handled
```

Leaving the handler alone is the right call and I am not suggesting changing it. uvicorn's `capture_signals` saves whatever handler it finds, restores it after the drain, and re-raises into it, so stomping on a bot's handler would change behavior that works today.

The consequence is worth naming, though. `signal.signal` replaces rather than chains, so a bot module (or an SDK it imports) that registers its own SIGTERM handler at import time evicts `_on_sigterm` for the rest of the startup window. From that point the window is only as safe as that handler. If it sets a flag and returns, which is a very ordinary shape, an early SIGTERM does nothing and the container starts and serves on a deleted pod, exactly as before this PR.

And tini does not backstop this case. Its `-e 143` remap only applies to a child that dies on the default disposition; once Python has any handler installed, the default disposition no longer applies, so the process survives. Both layers of the fix are bypassed by the same three lines of customer code.

The test that covers this is asserting the shape is intended, which I read as deliberate:

```python
# pipecat-base/tests/test_early_sigterm.py
def test_defer_keeps_a_handler_installed_by_the_bot_module():
    """defer_to_server() must not evict a handler the customer's import
    installed: uvicorn saves and restores it, and its post-drain re-raise
    invokes it, so that handler behaves exactly as it did before this module."""
```

What I would change is the changelog, which currently promises the fix without the qualification:

```markdown
# pipecat-base/CHANGELOG.md:10-13
- A stop signal that arrives while the agent is still starting up - anywhere
  in the imports, your `bot.py` included - now terminates the container.
```

One sentence would cover it, something like: "If your `bot.py` or one of its imports installs its own `SIGTERM` handler at import time, that handler replaces ours and is responsible for exiting."

To be clear about what I did and did not check: I did not enumerate which pipecat or SDK code paths install a SIGTERM handler at import time, so I do not know whether this is theoretical or something a real customer image already does. `PipelineRunner(handle_sigterm=True)` is not this case, since it runs at runtime and defaults to `False`. Do you have a sense of whether any of the starters or common transports do this?

cbrianhill (approved):
Checked the things I could verify from the repos I have:

- **The new ENTRYPOINT is what actually runs in production.** The operator builds the user container with no `Command` or `Args` (`pipecat-cloud-operator/internal/controller/pod_spec_builder.go:163-184`), so the image's own ENTRYPOINT/CMD are used verbatim and tini is really PID 1 on a real pod, not just under `docker run`.
- **`defer_to_server` covers the whole gap it claims to.** `server` is the module-level `WaitingServer` (`app.py:88`), and `defer_to_server` runs immediately before `server.run()` (`app.py:701`), so between deferral and uvicorn installing its own handler the predicate reads `False` and a SIGTERM still exits.
- **The absorb depends on stock uvicorn behavior, and nothing here overrides it.** `WaitingServer` overrides `shutdown` only (`waiting_server.py:35`); `capture_signals`, which saves the handler it finds, restores it, and then re-raises the captured signal, is untouched.
- **`pyproject.toml` is still at 0.1.27** while the changelog entry sits under `[Unreleased]`. That is finding 1.
- Could not verify the measured claims: the atexit hook never running under the `SIG_DFL` first cut, the two quickstart runs, and the emulated amd64 pass all happened outside this repo.

Two things worth calling out as done well. Gating `build-base` on the container check for `push` and not only `pull_request` is the part that actually protects customers, since the pull_request run does not push. And the `SIG_DFL` finding is the kind of thing that only turns up if you go looking: resetting to the default disposition looks obviously correct, and it silently skipped every atexit hook, which is where OpenTelemetry, Sentry and ddtrace flush.

Three findings, one of them a merge-order decision rather than a code defect:

1. Version not bumped, so merging republishes `dailyco/pipecat-base:0.1.27` with different runtime behavior (CHANGELOG.md:8).
2. The early-exit path leaves no record in the structured log lane, and drops captured import output (pcc_early_sigterm.py:65).
3. A bot module that installs its own SIGTERM handler at import time re-opens the window, and the changelog does not say so (pcc_early_sigterm.py:78).

Approving.

pipecat-base/CHANGELOG.md
mattshep: Agreed — bumped to 0.1.28 in 2d45e46 (`uv version` plus `uv lock`; the Dockerfile mounts the lock with `--locked`, so the refresh isn't optional), and the changelog section is now `[0.1.28] - 2026-09-09`.

Mark normally owns the version, so flagging in case he'd rather this were 0.2.0 given it changes what the container does on a signal. The README's separate-release-PR flow has this republish gap in general — every merge under `[Unreleased]` republishes the current tag — which is worth fixing on its own.

pipecat-base/pcc_early_sigterm.py
mattshep: Done in 2d45e46, in your lock-free shape: the handler appends a WARNING record to `$PCC_LOG_DIR/bot.jsonl` in `_serialize`'s exact form via `os.open`/`os.write` — opened at signal time rather than held from `install()`, since we run before `pcc_structured_logs` has created the directory.

It also sleeps one pump beat (50ms, the hop PCC-1038 documents) before `os._exit`, so the bot module's import output drains into the capture lane instead of being truncated at the moment you'd want it.

Verified inside the container with `PCC_LOG_DIR` set: the import trace and the record both land in `bot.jsonl`, in order. Unit test added (`test_early_exit_is_recorded_in_the_structured_lane`).

pipecat-base/pcc_early_sigterm.py
mattshep: Changelog sentence and module docstring in 2d45e46.

To your question, I checked: pipecat installs SIGTERM handlers only in `workers/runner.py`, at run time inside `run()`, gated by `handle_sigterm` (default `False`); the `pipecatcloud` SDK installs none; none of our bots do. Sentry, ddtrace and OpenTelemetry hook atexit/excepthook, not signals. So it's theoretical today, and it stays a documented contract rather than something to engineer around — wrapping a customer's own handler to force an exit would change semantics they chose.

PR #105, opened 2026-09-10
Give the serialization-failure marker a @timestamp; bump to 0.1.29 (PCC-1190)

mattshep:
The one record the structured log lane wrote without a @timestamp was its
own fallback — the marker emitted when a line cannot be serialised. Every
store so far indexed the record's own @timestamp field, so a missing one
was merely a missing field. The Cloud Logging store kind (PCC-1190) sets
each entry's time FROM @timestamp, and a record without one keeps the
Fluent Bit read time instead: after a restart with a backlog to drain that
is hours late, outside any time-range query for the session, and it does
not dedup across a checkpoint-less re-read either (the shipper's insertId
pairs with the timestamp). Found in review of the shipper's entry file
(sandbox#811); the queue-sidecar's twin fallback got the same fix there.

The marker now carries datetime.now().astimezone().isoformat() — the same
tz-aware shape pcc_early_sigterm.py writes and the lane's parsers accept.
The existing failure-path test asserts it. Version bumped to 0.1.29 so the
image publishes; the changelog entry is written for the people who rebuild
on it.


pipecat-base/pcc_structured_logs.py
```diff
@@ -160,7 +161,18 @@ def _serialize(record) -> str:
...
+        # otherwise file this line at the moment it was SHIPPED, hours late
+        # after a backlog restart and outside any time-range query, and a
+        # checkpoint-less re-read would not dedup it.
+        return json.dumps(
+            {
+                "@timestamp": datetime.now().astimezone().isoformat(),
+                "stream": "app",
+                "line": "<pcc_structured_logs: serialization failed>",
+            }
+        )
```
cbrianhill: **1. The marker still has no `session_id`, so a session-scoped query misses it for a second reason** — `suggestion`

`pipecat-base/pcc_structured_logs.py:169-175`

The timestamp fix closes the time axis. The session axis is still open, and the customer-facing log read filters on both.

```python
# pipecat-base/pcc_structured_logs.py:147-175
def _serialize(record) -> str:
    try:
        extra = record["extra"]
        stream = extra.get("pcc_stream", "app")
        payload = {"@timestamp": record["time"].isoformat(), "stream": stream}
        if stream == "app":
            payload["level"] = record["level"].name
            session_id = extra.get("session_id")
            if session_id and session_id != "NONE":
                payload["session_id"] = session_id
        else:
            if _current_session_id:
                payload["session_id"] = _current_session_id
        payload["line"] = record["message"]
        return json.dumps(payload, ensure_ascii=False, separators=(",", ":"))
    except Exception:
        # ... (comment elided)
        return json.dumps(
            {
                "@timestamp": datetime.now().astimezone().isoformat(),
                "stream": "app",
                "line": "<pcc_structured_logs: serialization failed>",
            }
        )
```

Session is a required clause in the query, not a hint. On the OpenSearch path the sandbox API pushes it into the `must` array alongside the `@timestamp` range:

```ts
// pipecat-cloud-sandbox api/src/services/logSource.ts:161-168
if (params.sessionId) {
  (query.body?.query?.bool?.must as QueryContainer[]).push({
    match_phrase: {
      log: params.sessionId,
    },
  });
}
```

The region-agent's own builder does the same with a `term`:

```go
// pipecat-cloud-region-channel internal/logquery/builder.go:35
must = append(must, map[string]any{term: map[string]any{"session_id": p.SessionID}})
```

And the Cloud Logging kind this ticket adds appends `AND labels.session_id="…"` to its filter string, since `session_id` is one of the keyword fields the shipper nests under `logging.googleapis.com/labels`. So on all three store kinds, a record with no `session_id` cannot come back from a session query no matter what its timestamp says. Fixing the time axis alone leaves the marker findable by a broad service-wide sweep but not by the per-session view a customer or an on-call engineer actually opens.

`_current_session_id` is reachable from the except block. It is a module global, so nothing about the malformed record can make it unreadable, and the raw lane already treats it as the authoritative session for a line that carries no loguru context. `run_bot` stacks the loguru context and the module slot together, so it is set for the whole session:

```python
# pipecat-base/app.py:276-278
        logger.contextualize(session_id=args.session_id),
        ...
        pcc_structured_logs.session_scope(args.session_id),
```

Suggested fix, keeping the schema's field order (`@timestamp`, `stream`, `session_id`, `line`):

```python
    except Exception:
        marker = {
            "@timestamp": datetime.now().astimezone().isoformat(),
            "stream": "app",
        }
        if _current_session_id:
            marker["session_id"] = _current_session_id
        marker["line"] = "<pcc_structured_logs: serialization failed>"
        return json.dumps(marker)
```

Best-effort by construction: a marker emitted outside any session still carries none, which is correct.

pipecat-base/pcc_structured_logs.py
```diff
@@ -160,7 +161,18 @@ def _serialize(record) -> str:
...
-        return json.dumps({"stream": "app", "line": "<pcc_structured_logs: serialization failed>"})
+        # It carries a @timestamp like every other record (PCC-1190): a store
+        # that stamps entries with the record time — Cloud Logging — would
+        # otherwise file this line at the moment it was SHIPPED, hours late
+        # after a backlog restart and outside any time-range query, and a
+        # checkpoint-less re-read would not dedup it.
+        return json.dumps(
+            {
+                "@timestamp": datetime.now().astimezone().isoformat(),
+                "stream": "app",
```
cbrianhill: **2. No `level`, so the marker lands at Cloud Logging severity `DEFAULT`** — `nit`

`pipecat-base/pcc_structured_logs.py:172`

The marker declares `"stream": "app"` but carries no `level`. Every other app-lane record has one, and the module docstring lists `level` as an app-lane field with no exception:

```
# pipecat-base/pcc_structured_logs.py:20-27 (docstring)
    @timestamp  RFC 3339 emit time
    stream      "app" (framework logger) | "stdout" | "stderr"
    level       loguru level name (app lane only; raw streams have no level)
    session_id  active session (omitted when none) ...
    line        the log line body
```

On the Cloud Logging kind the shipper copies `level` into the entry's severity, and the plugin's table maps an absent or unrecognised value to `DEFAULT`, which sorts below `DEBUG`. A console view or an alerting filter set to `severity >= WARNING` will not show the one record that says the log lane broke.

The failure path cannot know the original record's level — a missing `level` key is one of the things that raises here in the first place — but the marker's own level is not in doubt. A line that could not be serialised is an error:

```python
{
    "@timestamp": datetime.now().astimezone().isoformat(),
    "stream": "app",
    "level": "ERROR",
    "line": "<pcc_structured_logs: serialization failed>",
}
```

A nit rather than a suggestion because only the Cloud Logging kind gives `level` query meaning. The ES-family read paths never filter on it; the region-agent only carries it back as a response field (`internal/logquery/opensearch.go:84`), and the sandbox API never puts it in a query at all.

pipecat-base/tests/test_serializer.py
```diff
@@ -66,6 +66,13 @@ def test_serialization_failure_yields_marker_not_exception():
     # A record missing required keys must not raise.
     out = json.loads(_serialize({"extra": {}}))
     assert "serialization failed" in out["line"]
+    assert out["stream"] == "app"
+    # The marker is a record like any other (PCC-1190): a store that stamps
+    # entries with the record time — Cloud Logging — files it when it was
+    # emitted, not when it was shipped, which can be hours apart after a
+    # backlog restart. Same tz-aware isoformat as every other record.
+    ts = datetime.fromisoformat(out["@timestamp"])
+    assert ts.tzinfo is not None
```
cbrianhill: **3. The test asserts tz-awareness, which is weaker than the property the fix depends on** — `suggestion`

`pipecat-base/tests/test_serializer.py:74-75`

```python
# pipecat-base/tests/test_serializer.py:65-75
def test_serialization_failure_yields_marker_not_exception():
    # A record missing required keys must not raise.
    out = json.loads(_serialize({"extra": {}}))
    assert "serialization failed" in out["line"]
    assert out["stream"] == "app"
    # ... (comment elided)
    ts = datetime.fromisoformat(out["@timestamp"])
    assert ts.tzinfo is not None
```

Nothing downstream uses `datetime.fromisoformat`. Fluent Bit's parser filter tries three regexes in order and copies the parsed time into the record time only if one of them both matches and yields a time:

```
# pipecat-cloud-sandbox pipecat-log-shipper/parsers.conf (PR #811)
[PARSER]
    name        pcc_rfc3339
    regex       ^(?<time>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d+[+-]\d{2}:\d{2})$
    time_format %Y-%m-%dT%H:%M:%S.%L%z

[PARSER]
    name        pcc_rfc3339_nofrac
    regex       ^(?<time>\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}[+-]\d{2}:\d{2})$
    time_format %Y-%m-%dT%H:%M:%S%z

[PARSER]
    name        pcc_rfc3339_other
    regex       ^(?<time>.+)$
    time_format %Y-%m-%dT%H:%M:%S.%L%z
```

`datetime.fromisoformat` is much more permissive than either strict regex. On 3.11+ it accepts a `Z` suffix, a bare date, a space instead of `T`, and an offset carrying seconds. Every one of those satisfies `assert ts.tzinfo is not None` (or, for a bare date, fails on a different line than the one intended) while falling through to `pcc_rfc3339_other`, the parser whose entry-file comment says it exists to be loud about shapes no emitter writes. A future edit to `datetime.now().isoformat() + "Z"` would keep this test green and silently move the marker onto the catch-all.

The value today is correct, so this is about what the test protects, not about a bug in the change.

Suggested fix — assert the wire shape the shipper matches, which is also how the rest of this file reads (`test_app_lane_fields` asserts the exact `@timestamp` string):

```python
import re

...
    assert re.fullmatch(
        r"\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?[+-]\d{2}:\d{2}",
        out["@timestamp"],
    )
```

The optional fraction covers `isoformat()`'s habit of dropping it when the microsecond is exactly zero, which is the case `pcc_rfc3339_nofrac` exists for.

pipecat-base/pcc_structured_logs.py
```diff
@@ -160,7 +161,18 @@ def _serialize(record) -> str:
...
         # Never let serialization break logging; emit a marker record instead.
-        return json.dumps({"stream": "app", "line": "<pcc_structured_logs: serialization failed>"})
+        # It carries a @timestamp like every other record (PCC-1190): a store
+        # that stamps entries with the record time — Cloud Logging — would
+        # otherwise file this line at the moment it was SHIPPED, hours late
+        # after a backlog restart and outside any time-range query, and a
+        # checkpoint-less re-read would not dedup it.
+        return json.dumps(
+            {
+                "@timestamp": datetime.now().astimezone().isoformat(),
```
cbrianhill: **4. Local offset rather than forced UTC is the right call here** — `praise`

`pipecat-base/pcc_structured_logs.py:171`

Worth recording, because the twin fix in the queue sidecar went the other way:

```python
# pipecat-cloud-sandbox pipecat-queue-sidecar/structured_logs.py (PR #811)
"@timestamp": datetime.now(tz=timezone.utc).isoformat(),
```

In this file the normal path stamps `record["time"].isoformat()`, and loguru's record time is local and tz-aware. `pcc_early_sigterm.py:88` already writes `datetime.now().astimezone().isoformat()`. A UTC-forced marker would have been the one record in `bot.jsonl` whose offset differed from its neighbours in any image that sets `TZ`, which customer bots built FROM this base are free to do. Both forms parse: the shipper's strict regexes take any `[+-]HH:MM`, so matching the file it lives in costs nothing downstream.

One thing to check on the other PR rather than here: `parsers.conf`'s comment says the emitters write "microseconds and a `+00:00` offset". That is true of the sidecar unconditionally and true of this file only when the container's timezone is UTC. The regexes are already written for the general case, so it is the comment that is narrow, not the code.

cbrianhill (approved):
Checked the claims I could verify from the repos:

- The parsers this has to satisfy accept a non-UTC offset. `pcc_rfc3339` and `pcc_rfc3339_nofrac` (sandbox#811, `pipecat-log-shipper/parsers.conf`) both match `[+-]\d{2}:\d{2}`, so `astimezone()` output parses wherever the container's `TZ` lands, and the no-fraction case that `isoformat()` produces at microsecond zero has its own parser.
- The shape matches what this repo already writes. `pcc_early_sigterm.py:88` is the identical `datetime.now().astimezone().isoformat()`, and the normal `_serialize` path uses loguru's local, tz-aware `record["time"]`.
- The version bump is what publishes. `docker-publish.yml` reads `.project.version` from `pipecat-base/pyproject.toml` for both the `0.1.29` and `latest` tags, and `versions.yaml` carries only the Python matrix, so nothing else needed touching. No stale `0.1.28` anywhere else in the tree.
- The test imports `datetime` already (`test_serializer.py:2`), so the new `datetime.fromisoformat` call is not a NameError waiting to happen. All eight checks are green on `24c52a2`.

The fallback stays as safe as it was: a fixed-key dict and a `datetime.now()` that cannot itself raise, so nothing new can escape the except block that exists to keep logging alive.

Three findings, none blocking:

1. The marker still has no `session_id`, so a session-scoped query misses it for a second reason (pcc_structured_logs.py:169-175) — suggestion.
2. No `level`, so it lands at Cloud Logging severity `DEFAULT`, under a `severity >= WARNING` view (pcc_structured_logs.py:172) — nit.
3. The test asserts tz-awareness, which is weaker than the wire shape the shipper's regexes require (tests/test_serializer.py:74-75) — suggestion.

Approved. The change does what it says and the fallback is no riskier than before. Finding 1 is the one I would actually take: it is the same class of gap on the other axis of the same query, and `_current_session_id` is already in scope in that except block.

pipecat-base/pcc_structured_logs.py
mattshep: Taken. The marker stamps `_current_session_id` when one is set — the same slot the raw lane reads — in the normal field order, and carries none outside a session. A test under `session_scope` asserts it, and the no-session case asserts it's absent. efa4e1c

pipecat-base/pcc_structured_logs.py
mattshep: Taken — `ERROR`. The record's own level is one of the things that may have raised, but a line that couldn't be serialised is an error, and DEFAULT sorts below DEBUG on the one store that gives level query meaning. efa4e1c

pipecat-base/tests/test_serializer.py
mattshep: Taken. The assertion is now the regex the shipper's strict parsers match, fraction optional, as a module constant whose comment says why `fromisoformat` isn't it (a Z suffix, a bare date and a space separator all parse there and would fall through to the catch-all). efa4e1c

pipecat-base/pcc_structured_logs.py
mattshep: Agreed on the `parsers.conf` comment: the regexes take any offset and the prose says `+00:00`. That's comment-only in a merged PR, so it rides along with the next change in that repo rather than getting a PR of its own.

PR #106, opened 2026-09-20
Publish what the observers report (PCC-656)

markbackman:
Builds on @sunahsuh's #91 — her two commits are cherry-picked here, so the transport is hers and the records are new. #91 can be closed in favour of this.

## What

Every record the observers report is published to the event pipeline as its own event, when `PIPECAT_EVENT_PUBLISHER_ENDPOINT` is set. Unset → nothing is published, exactly as before.

| `event_name` | from |
|---|---|
| `startup_timing`, `transport_timing` | `StartupTimingObserver` |
| `user_bot_latency`, `first_bot_speech_latency`, `latency_breakdown` | `UserBotLatencyObserver` |
| `service_latency`, `service_usage` | `ServiceMetricsObserver` |
| `speech_event` | `SpeakingObserver` |
| `function_call_event` | `FunctionCallObserver` |
| `error` | `ErrorObserver` |

An event's `event_properties` are the record's own fields (`model_dump(mode="json", exclude_none=True)`), so a record that gains a field in the framework gains it in the warehouse without a change here.

What #91 published is unchanged in spirit but no longer in shape. `latency_breakdown` carried `chronological_events()` — a list of pre-rendered strings like `"OpenAILLMService#0: TTFB 0.336s"`, which can't be grouped or summed in SQL and only covered the services that reported a metric. It now carries `contributions`, which name every part of the turn, sum to `total_secs`, and each have a stable `key` and an `owner_kind`. `startup_timing` picks up the setup/start phase split added since #91. The four other observers are new since #91 and bring what nothing else reports — most notably LLM token counts, which is what cost analysis runs on.

## The old log lines are gone

`pcc_observers.py` no longer writes its `[pcc-observability]` lines for startup timing and turn latency — that data is published instead. Nothing read them: the dashboard's session timeline (`parseSessionTimeline.ts`) is built from Pipecat's own log output (`"User started speaking"`, `"OpenAILLMService#0 TTFB: 1.181s"`, `"prompt tokens: 57, completion tokens: 33"`), and no repo references the `[pcc-observability]` prefix.

**One consequence to weigh:** with `PIPECAT_EVENT_PUBLISHER_ENDPOINT` unset, this data now goes nowhere. That is every deployment until the publisher is rolled out beyond staging `us-west-2`, plus anyone running the base image outside Pipecat Cloud. If we would rather not lose that window, the lines can come back until the publisher is deployed everywhere.

## Version tolerance

The base image doesn't pin Pipecat; the bot's image does. Each observer is imported on its own and skipped when the import fails, so a bot on an older Pipecat reports what its version has rather than failing to start. Service metrics, speaking, function calls and errors need Pipecat **1.11.0**; startup timing and latency work as before.

## Two guards beyond #91

- **`_events_url()` completes a bare host and port to `/events`.** The publisher routes `POST /events` and 404s everything else, and a publish failure is swallowed as a warning — so an endpoint configured without the path would drop every event while looking like a quiet session. Happy to delete this and put the full path in the env var instead; it exists because the staging endpoint was handed to me as bare host and port.
- **`_publish_record()` drops a record that won't serialize.** A function call's result is `Any` — whatever the customer's handler returned — so the dump can raise inside the handler that reported it.

## Privacy call worth a second opinion

`FunctionCallObserver(include_arguments=False, include_results=False)`. Tool arguments are model-generated from the conversation and can carry names, addresses, account numbers. Writing them into our warehouse reads like the customer's decision rather than ours, so both are off. Trivial to flip if we disagree.

## Open questions, deliberately not changed here

- **`session_id` is the literal `"NONE"` outside a session.** #91 does this, and I've kept it. `pcc_structured_logs` treats `"NONE"` as a placeholder worth omitting; the publisher requires a non-empty string, so the options are a `"NONE"` row or no row. Worth deciding which we want in BigQuery.
- **Nothing bounds in-flight requests.** Handlers run as tasks, so a slow endpoint never blocks the pipeline — but a hung one would accumulate pending tasks with 5s timeouts for a session's length. The publisher rate-limits at 100/s per pod and we emit ~1–3/s, so this is about the endpoint being unhealthy, not about volume. A semaphore and a shorter timeout would bound it.
- **No flush at session end.** Records emitted in a session's final moments can be in flight when the pod exits — which are the records you most want when investigating a bad call. `app.py`'s `session_scope` is the seam if we want one.

## Testing

`uv run pytest` — 59 pass, 11 new. The new tests pin the envelope against the publisher's own validator (the ISO-8601 regex, the 256-char field cap and the 10KB properties cap are copied from `types.ts`), so a contract change fails here rather than going quiet in production.

The observer wiring is not unit-tested: the repo's tooling declares `requires-python = ">=3.10"` and Pipecat 1.11 needs 3.11, so importing the new observers in tests would mean raising that floor. The handlers are ten near-identical lines and the staging deploy covers them.

**Not yet verified end-to-end.** The publisher identifies its caller by pod IP, so it can't be reached from a laptop. Next step is building this image, deploying to staging `us-west-2`, and confirming rows land in `daily-data-platform-staging.live_ingestion.pipecat_cloud_events`.


mattshep (changes requested):
## Overview

Reviewed at `83beaf0`.

Publishes every observer record the base image attaches to the event pipeline, replacing the hand-built payloads and the observability log lines. Since the description already says not to merge yet, I've focused on what should change before it does: what leaves the customer's bot, and which session it's attributed to.

The transport is sound. Two things need changing.

## Verification

**Publishing can't stall a bot.** The observers register their events without `sync`, so Pipecat runs each handler call as its own task. A publisher that hangs, which a dropped connection would, costs a pending task for up to the 5-second timeout and never holds up the pipeline. Failures are logged without the record.

**Version guards work as described.** Each observer is imported on its own, and the tests hide modules to prove an older Pipecat still starts with what it has.

**Speech, latency, timing and usage records carry no conversation content.** A speech event is a kind and two timestamps. Usage is counts, the model and the processor name. Those are fine to publish.

## Findings

**Finding 1 (please address): error text publishes the content the PR says it withholds, and nothing stops the next field that does.**

Arguments and results are withheld on the principle that "what a bot's tools were asked and answered belongs to the bot". Two fields carry exactly that, by another route:

- **`FunctionCallEvent.error`** is the tool handler's exception text, and the include flags don't cover it. A tool's error usually quotes what it was asked: "No reservation for Jane Smith on the 30th" is an argument and a result in one string.
- **`ErrorEvent.message`** is the failing processor's own words. A provider error can echo the request it rejected, and a custom processor's message can hold anything.

The staging check found no arguments or results, but only because that tool failed with a message that held none.

The underlying cause is publishing `model_dump()` of every record. The description presents "a field added in the framework reaches the warehouse without a change here" as a benefit, but for data that leaves a customer's bot it's the wrong default. A field that starts carrying conversation content, a transcript on a speech event say, would ship with no one here deciding it should.

I'd publish an explicit list of fields per record type, so a new field is withheld until someone chooses to add it. For errors, `exception_type`, `category`, `processor` and `processor_usable` are what grouping needs; `ErrorEvent`'s own docs say failures group by exception type because the message carries specifics. The message text should stay with the bot, the same as arguments and results.

**Finding 2 (please address): the session id is a second copy of a slot that already exists, and it's never cleared.**

`pcc_structured_logs` already keeps the current session. `session_scope` sets it in `run_bot`, and its `finally` clears it on every exit path. This PR adds `GLOBALS["current_session_id"]`, set on the line before `session_scope` is entered and never cleared. So on a reused pod, anything published after a session ends is attributed to that session rather than to none. `"NONE"` only ever appears before the first session the process runs.

Reading the existing slot fixes attribution, and it removes the need for the new `shared_state` module. It also settles the `"NONE"` question consistently. Outside a session the slot is empty, and since every observer is attached inside one, anything published then is a bug rather than a real session. I'd skip publishing it rather than write a row that can't be joined to anything.

## On the questions raised in the description

- **Function call arguments and results off:** agreed, and Finding 1 extends the same reasoning to error text.
- **Completing a bare endpoint to `/events`:** I'd drop it. The platform sets this variable, so the fix belongs where the value is set, and a rewrite makes the variable mean two different things.
- **`"NONE"` outside a session:** covered in Finding 2.

## Verdict

Requesting changes for Findings 1 and 2, which also matches the description's own note not to merge yet. The transport, the version guards and the tests against the publisher's contract are right, and shouldn't change.

markbackman:
Thanks — both findings were right, and both are fixed. Pushed as three commits on top of what you reviewed, so your comments still point at `83beaf0`.

**Finding 1 — error text.** You're right that the flags didn't cover it, and that `model_dump()` was the real problem rather than those two fields. There's now a `_PUBLISHED_FIELDS` list naming what each record publishes. Withheld: a tool's `arguments`, `result` and the exception its handler raised, plus `ErrorEvent.message`. A failure still publishes its type, category, processor and `processor_usable`, and `push_error_frame` logs the message for the bot's own session, so the detail is still there for whoever can see that customer's logs — it just doesn't go to the warehouse.

I stopped at the top level rather than going into nested structures. `contributions`, `ttfb` and `processor_timings` travel whole — they hold processor names, span keys and durations that the framework writes. That's the one place a new field could still arrive unreviewed, so say if you'd rather it were enumerated too.

**Finding 2 — the session slot.** Fixed as you described. `_publish_event` reads `pcc_structured_logs.current_session()` (a getter I added next to `set_current_session`), `shared_state.py` is deleted, and `app.py` and the Dockerfile are back to byte-identical with main. Nothing is published outside a session, so `"NONE"` is gone.

One consequence worth naming: a record emitted during the 250ms linger after a session ends still carries that session. That's what the linger is for in the log lane and it seems right here too — a bot's last records belong to the session that produced them.

**`/events`.** Dropped, and you were right about the reason I gave. I'd justified it as guarding a silent failure, but a wrong URL logs a 404 for every record it loses, so it isn't silent at all. The catch is that the staging secret currently holds the bare host and port, so that value needs `/events` before this lands or our test agent goes quiet.

**One thing I couldn't do:** a test asserting every published name still exists on its model. The list names fields across Pipecat versions — `contributions` and `measured_from` don't exist in the 1.6.0 this repo installs — so that test would fail against a correct list. Making it work means raising the tooling floor to Python 3.11 so the repo can import 1.11's models. Your call whether that's worth it; the drift it would catch loses a column rather than leaking one. In the meantime the tests hold the policy itself: the content-bearing fields stay out, every record named at a call site has a list, and a record publishes only what its list names.

mattshep (approved):
## Re-review: `44b72d4`

Approving. Both findings are fixed, and I checked each against the commits and the framework models.

## Finding 1: published fields

`_PUBLISHED_FIELDS` names what each record publishes, and `model_dump(include=...)` enforces it, so a field added upstream stays with the bot until someone adds it here. The content-bearing fields are out: a tool's arguments, result and exception text, and an error's message. A failure still publishes its type, category and processor, which is what grouping needs.

**Naming fields across Pipecat versions is safe.** Pydantic ignores an `include` name the installed model doesn't have. I checked on 2.13, so a list that names 1.11 fields still publishes what an older version has.

**The tests hold the policy rather than one instance of it.** The content fields are asserted out, every record named at a call site has a list, and a model with an extra field publishes only what its list names. That last one uses a `transcript` field, which is exactly the case this is for.

**On the nested structures you asked about:** I read every nested model. `ttfb`, `contributions`, `text_aggregation` and `function_calls` hold processor names, keys, owners, function names and timings, and nothing written by the conversation. So nothing leaks today.

I'd still enumerate them in this PR. It's the one place the rule you've just written, withheld until someone adds it, doesn't yet hold. `text_aggregation` is the plausible first place for a text field to appear, since it measures aggregating text. Pydantic's `include` takes nested dicts, and `{"__all__": {...}}` for lists, so it's a few lines, and cheaper now than after something has leaked.

**On the test that published names exist on their models:** I wouldn't raise the Python floor for it. The list is cross-version by design, and as you say, the drift it would catch loses a column rather than leaking one.

## Finding 2: session attribution

`_publish_event` reads `pcc_structured_logs.current_session()`. `session_scope` is entered in `run_bot` whether or not structured logs are enabled, so the slot is always set during a session. A new session cancels the linger before taking the slot, so attribution can't bleed into the next session, and records in the 250ms after one ends belonging to it is right. Nothing is published outside a session, and `shared_state` is gone.

**The `/events` completion** is dropped. The staging value needs `/events` appended before this lands, as you noted.

## Verdict

Approve. The fields that can quote a conversation stay with the bot, and records are attributed to the session that produced them. Enumerating the nested structures is the one thing I'd still do here.

This clears my review, but the conditions in the description still hold: the publisher's availability, the egress exception, and the staging value.

pipecat-base/tests/test_event_publisher.py
```diff
@@ -0,0 +1,284 @@
...
+    """Telemetry that cannot be delivered must not disturb the bot."""
+
+    class _Broken:
+        closed = False
+
+        def post(self, url, json=None):
+            raise OSError("connection refused")
+
+    monkeypatch.setattr(pcc_observers, "_get_http_session", lambda: _Broken())
+    monkeypatch.setattr(pcc_observers, "_events_endpoint", "http://publisher:3000/events")
```
sunahsuh: This test never reaches `_Broken.post`. It sets the endpoint but not the session slot, and `_current_session_id` defaults to `None`, so `_publish_event` returns before posting. I confirmed by counting calls with this setup: zero as written, one once a session is set.

Set the slot as the `posted` fixture does (or reuse it with the broken session), and assert the post was attempted, so the test actually proves the exception is swallowed.

pipecat-base/pcc_observers.py
```diff
@@ -2,21 +2,178 @@
...
+    """
+    if not _events_endpoint:
+        return
+
+    session_id = pcc_structured_logs.current_session()
+    if not session_id:
+        return
+
+    payload = {
+        "ts": datetime.now(timezone.utc).isoformat(timespec="milliseconds"),
```
sunahsuh: `timespec="microseconds"` would remove the ordering caveat in the description entirely. The publisher's regex accepts any number of fractional digits (`(\.\d+)?`), and every record already carries a microsecond `timestamp`, so there's no reason for the envelope to be coarser.

pipecat-base/pcc_observers.py
```diff
@@ -74,18 +213,83 @@ async def _setup_user_bot_latency_observer(worker):
 
     @observer.event_handler("on_latency_measured")
     async def on_latency_measured(observer, latency_seconds):
-        logger.info(f"[pcc-observability] User-bot latency | latency={latency_seconds:.3f}s")
+        await _publish_event("user_bot_latency", {"latency_secs": round(latency_seconds, 3)})
```
sunahsuh: `user_bot_latency` and `first_bot_speech_latency` are the only two records with no `timestamp` of their own, so the consumer note ("order by the record's `timestamp`, not the envelope's `ts`") has nothing to use for them. Adding `"timestamp": time.time()` to both payloads would make them consistent with the other eight.

pipecat-base/pcc_observers.py
```diff
@@ -2,21 +2,178 @@
...
+# POST /events and 404s anything else.
+_events_endpoint = environ.get("PIPECAT_EVENT_PUBLISHER_ENDPOINT")
+_http_session: aiohttp.ClientSession | None = None
+
+
+def _get_http_session() -> aiohttp.ClientSession:
+    global _http_session
+    if _http_session is None or _http_session.closed:
+        _http_session = aiohttp.ClientSession(
+            timeout=aiohttp.ClientTimeout(total=5),
```
sunahsuh: Worth thinking about the state the description says staging is in today: endpoint set, egress exception not yet in place. Every record then hangs for the full 5s and logs one warning, so a session produces a couple of hundred warnings, and the observer's `cleanup()` waits on the stragglers at session end. It never stalls the pipeline, but a shorter `connect=` timeout, or a "stop trying after N consecutive failures" guard, would make a misconfigured region cheap instead of noisy.

pipecat-base/pcc_observers.py
```diff
@@ -2,21 +2,178 @@
...
+        "start_time": True,
+        "bot_connected_secs": True,
+        "client_connected_secs": True,
+    },
+    "latency_breakdown": {
+        "measured_from": True,
+        "total_secs": True,
+        "user_turn_start_time": True,
+        "user_turn_secs": True,
+        "contributions": True,
```
sunahsuh: Seconding Matt's note on enumerating the nested structures. Pydantic's `include` takes `{"__all__": {...}}` for lists, so it's a few lines, and it closes the one place the "withheld until someone adds it" rule doesn't yet hold. Not blocking.

pipecat-base/pcc_observers.py
```diff
@@ -2,21 +2,178 @@
...
+from os import environ
 
+import aiohttp
+import pcc_structured_logs
 from loguru import logger
 
+# The full URL records are posted to, route included: the publisher serves
+# POST /events and 404s anything else.
+_events_endpoint = environ.get("PIPECAT_EVENT_PUBLISHER_ENDPOINT")
+_http_session: aiohttp.ClientSession | None = None
```
sunahsuh: Minor: the client session is never closed, so the process will emit an "Unclosed client session" warning at exit. Either a shutdown hook or a comment saying it's intentional.

pipecat-base/pcc_observers.py
```diff
@@ -2,21 +2,178 @@
...
+bot pinned to an older Pipecat reports whatever its version has rather than
+failing to start.
 """
 
-import json
+import uuid
+from datetime import datetime, timezone
+from os import environ
 
+import aiohttp
```
sunahsuh: Minor: `aiohttp` isn't a declared dependency of pipecat-base. It arrives through pipecat-ai and pipecatcloud, which is safe because this file only loads under pipecat, but a one-line comment saying so would save the next reader a search.

pipecat-base/CHANGELOG.md
```diff
@@ -5,6 +5,32 @@ All notable changes to the **Pipecat Cloud Base Images** will be documented in t
...
+  Function call arguments and results are not published: what a bot's tools
+  were asked and answered belongs to the bot, not to the platform running it.
+
+### Removed
+
+- The `[pcc-observability]` log lines reporting startup timing and turn
+  latency. What they reported is published as events instead, and a session's
+  timeline in the dashboard is built from Pipecat's own log output rather than
+  from these.
 ## [0.1.31] - 2026-09-22
```
sunahsuh: Nit: missing a blank line before this heading.

pipecat-base/tests/test_event_publisher.py
sunahsuh: This test never reaches `_Broken.post`. It sets the endpoint but not the session slot, and `_current_session_id` defaults to `None`, so `_publish_event` returns before posting. I confirmed by counting calls with this setup: zero as written, one once a session is set.

Set the slot as the `posted` fixture does (or reuse it with the broken session), and assert the post was attempted, so the test actually proves the exception is swallowed.

pipecat-base/pcc_observers.py
sunahsuh: `timespec="microseconds"` would remove the ordering caveat in the description entirely. The publisher's regex accepts any number of fractional digits (`(\.\d+)?`), and every record already carries a microsecond `timestamp`, so there's no reason for the envelope to be coarser.

I checked that the extra precision survives the whole path: the publisher returns `ts` untouched, the Pub/Sub schema carries it as a string, the BigQuery subscription uses the topic schema, and the column is a `TIMESTAMP` (microsecond precision). A literal `TIMESTAMP("2026-09-23T12:15:41.187123+00:00")` parses to the microsecond in the staging project, and every row in `live_ingestion.pipecat_cloud_events` today has a `.xxx000` microsecond component, so the millisecond value is exactly what is arriving. Nine fractional digits are rejected, so keep it at six.

pipecat-base/pcc_observers.py
sunahsuh: `user_bot_latency` and `first_bot_speech_latency` are the only two records with no `timestamp` of their own, so the consumer note ("order by the record's `timestamp`, not the envelope's `ts`") has nothing to use for them. Adding `"timestamp": time.time()` to both payloads would make them consistent with the other eight.

pipecat-base/pcc_observers.py
sunahsuh: Worth thinking about the state the description says staging is in today: endpoint set, egress exception not yet in place. Every record then hangs for the full 5s and logs one warning, so a session produces a couple of hundred warnings, and the observer's `cleanup()` waits on the stragglers at session end. It never stalls the pipeline, but a shorter `connect=` timeout, or a "stop trying after N consecutive failures" guard, would make a misconfigured region cheap instead of noisy.

pipecat-base/pcc_observers.py
sunahsuh: Seconding Matt's note on enumerating the nested structures. Pydantic's `include` takes `{"__all__": {...}}` for lists, so it's a few lines, and it closes the one place the "withheld until someone adds it" rule doesn't yet hold. Not blocking.

pipecat-base/pcc_observers.py
sunahsuh: Minor: the client session is never closed, so the process will emit an "Unclosed client session" warning at exit. Either a shutdown hook or a comment saying it's intentional.

pipecat-base/pcc_observers.py
sunahsuh: Minor: `aiohttp` isn't a declared dependency of pipecat-base. It arrives through pipecat-ai and pipecatcloud, which is safe because this file only loads under pipecat, but a one-line comment saying so would save the next reader a search.

pipecat-base/CHANGELOG.md
sunahsuh: Nit: missing a blank line before this heading.

sunahsuh (approved):
Overall looks good to me! Claude had a few additional comments that I think are worth considering but definitely nothing blocking – I think particularly the lack of timestamp for those two events might cause issues down the line

pipecat-base/tests/test_event_publisher.py
markbackman: Fixed. Good catch — it was passing for the wrong reason. It now sets the session, reaches the post, and asserts the attempt happened; I checked it fails if the `except` is removed.

pipecat-base/pcc_observers.py
markbackman: Fixed. `timespec="microseconds"` now, which removes the ordering caveat from the description rather than documenting it. Thanks for tracing it all the way to the column.

pipecat-base/pcc_observers.py
markbackman: Fixed. Both now carry `"timestamp": time.time()`, so all ten records have their own time instead of two being special.

pipecat-base/pcc_observers.py
markbackman: Fixed. A session that fails ten times in a row stops trying and says so once. The count clears on a success and starts over each session, so a publisher that comes back is picked up without restarting the pod.

I left the timeout alone — the guard is what fixes the count you described, and a shorter `connect=` would also change behaviour for healthy regions.

pipecat-base/pcc_observers.py
markbackman: Fixed. All four nested structures enumerate their fields now, so a field added inside a contribution or TTFB entry is withheld like any other.

pipecat-base/pcc_observers.py
markbackman: Fixed, with a comment. The session lives as long as the process on purpose, to keep the connection open across a pod's sessions, so the warning at exit is the cost of not carrying a shutdown hook for a socket the kernel reclaims anyway.

pipecat-base/pcc_observers.py
markbackman: Fixed, with a comment — and the answer turned out simpler than either of us had it. `pipecat-ai` is not in pipecat-base's lockfile at all; `aiohttp` comes from `pipecatcloud`, which is a declared dependency, so it is there whether or not pipecat is.

pipecat-base/CHANGELOG.md
markbackman: Fixed.

markbackman:
This is updated and ready to be merged once production is updated with the endpoint and secrets.

mattshep (commented):
> [!WARNING]
> **Since I approved this at `44b72d4`, it has started publishing conversation transcripts: the full text of every user and assistant turn, for every deployment that isn't explicitly excluded.** My approval was of the version without them. I'm not blocking, and if this has already been decided, say so and I'll take it as settled. But it's a data-policy change, not a code one, and it should be merged on purpose rather than under an approval that didn't cover it.

## Transcripts are published by default

`f4aab8cf` publishes a `transcript` record per turn, and `2154ed8c` withholds it only where the platform sets `PCC_EXCLUDE_CONTENT`. So every other deployment's end-user conversations go to the event pipeline by default.

That's the reverse of what this PR said until now, and of what its CHANGELOG still says: "what a bot's tools were asked and answered belongs to the bot, not to the platform running it." A tool's arguments are withheld, but the conversation that produced them is published.

Before this merges, I'd want someone to own these questions explicitly:

- **Opt-in or opt-out?** Is publishing transcripts something a customer turns on, or something they have to turn off?
- **Retention and access:** how long transcripts are kept, and who can read them.
- **Sign-off:** whoever owns how we handle customer data should agree to the default, not just the code.

## The flag fails open

The comment says "anything but `false` excludes", but the check is `not in ("", "false")`, so an unset or empty `PCC_EXCLUDE_CONTENT` publishes. `2154ed8c`'s own reasoning, that a transcript published by mistake can't be recalled, argues for the opposite. Make `"false"` the only value that publishes, so an agent that never received the flag withholds transcripts rather than sending them.

## Exclusion is applied at deploy time

The platform sets this flag when an agent is deployed. Turning exclusion on for an organization doesn't reach agents that are already running, so their transcripts keep flowing until each one is redeployed. I've raised that where the flag is set.

## Everything else

The hardening since my approval is right: the nested fields are now listed too, publishing stops after repeated failures, and timestamps are at the warehouse's precision. None of this is about those commits.

markbackman:
Thanks — all three addressed.

**Transcripts by default.** Decided, not drifted: conversations already go to the application logs, and this puts them in a form a UI can read. @mark owns that call and has taken it up with the team; the retention and access questions sit with the platform rather than this PR.

**The flag failing open.** Fixed in `16a51fa` — only an explicit `"false"` publishes now, so an agent that never received `PCC_EXCLUDE_CONTENT` withholds. That also matches what the operator does with the field it comes from (`pod_spec_builder.go`: an unset `excludeContent` excludes), so both ends now read an absent value the same way. You were right that my own reasoning argued for this and the code did the opposite.

**Applied at deploy time.** Confirmed in the operator thread: agents only pick up the endpoint on redeploy, and that same redeploy stamps the flag — so there shouldn't be a window where the endpoint is set and the flag isn't. Turning exclusion on for an org still needs a redeploy to reach running agents, which matches how log exclusion behaves today.

One thing that follows for testing: our own test agent sets the endpoint by hand and has no `PCC_EXCLUDE_CONTENT`, so it now withholds transcripts until the control plane writes the field. Next round of testing goes to staging, where the auto-set path is live, rather than through the hand-configured secret.

PR #107, opened 2026-09-21
Run the flow a session names, not only the one the image ships (PCC-1241)

mattshep:
A Pipecat Flows bot whose flow is data can only run the flow baked into its
image, because nothing in a session carries one. Pipecat Cloud can now send the
flow a session named, and this is the image side of that: the flow reaches the
bot as `runner_args.flow_config`, which is where pipecat's own development
runner puts it (pipecat-ai/pipecat#5831), so one bot file reads the same field
running locally and running here.

A session that names a flow arrives as an envelope — the session body beside the
flow, marked by `X-Pcc-Start-Envelope`. The marker is what decides whether a
body is an envelope, rather than the shape of the JSON: a session body that
itself holds `body` and `flow_config` keys is a plausible thing for a caller to
send, and inferring would take their own data apart. Without the marker the
request body is the session body and nothing else, which is every request sent
until now, so nothing already running changes.

The flow is assigned to the session arguments rather than passed to their
constructor. It arrives on pipecat's `RunnerArguments` only from the release
that adds it, and the pipecat-ai version is the image author's choice rather
than ours; a constructor keyword would fail outright on an older one and take
the session with it. Assigning afterwards works on both, so a bot written
against the newer release reads the field either way, and a bot on an older
pipecat-ai still starts and simply ignores it.

The SmallWebRTC path needed the treatment the session body already has. There
the `/bot` request only holds the connection open and the bot runs from a later
`/api/offer` request that never saw these headers, so the flow waits beside the
body and is cleared with it when the session ends — nothing carries into the
next session on a reused pod.

A request marked as an envelope that cannot be read as one is refused with a 400
rather than started. The alternatives are running the bot on a body its caller
never sent, or on the flow the image ships with while the caller believes theirs
was accepted.

Version bumped to 0.1.30 so the image publishes, and the changelog entry is
written for the people who rebuild on it. The platform side (sandbox#843) sends
a flow only to an agent on 0.1.30 or newer, so an agent on an earlier image is
unaffected and its sessions keep running as they do today — the two versions
have to name each other, and that side's constant is set to this one.

`ruff format`, `ruff check` and `pytest` pass: 62 tests, 14 of them new. They
cover the envelope split, both refusals, the SmallWebRTC stash, and the case the
marker exists for — a body that merely looks like an envelope, with no flow
named, passing through untouched.


pipecat-base/app.py
```diff
@@ -135,6 +135,20 @@ def filter(self, record: logging.LogRecord) -> bool:
 _BUDGET_KEY = "pcc_session_budget"
 _warned_no_budget = False
 
+# A session may name the Pipecat Flows config it runs. Pipecat Cloud then sends
```
cbrianhill: **What this PR does**

A Pipecat Cloud session can now carry its own Pipecat Flows config, so one deployed agent can run a different conversation flow per session without rebuilding the image. The bot reads it as `runner_args.flow_config`, which is the same field pipecat's local development runner uses.

**Why:** before this, a Flows bot could only run the flow baked into its image. PCC-1241 adds per-session flows. The platform half is sandbox#843, and it only sends a flow to agents on base image 0.1.30 or newer.

**How:**
- `app.py` `/bot` handler: when the `X-Pcc-Start-Envelope: 1` header is present, the request body is read as `{"body": ..., "flow_config": ...}` and split by `_split_start_envelope`. Without the header, the body is used as-is, as before. A marked request that is not a valid envelope gets a 400.
- `_attach_flow_config` sets `flow_config` on the session arguments after they are constructed, so the assignment works even with an older pipecat-ai whose `RunnerArguments` has no such field.
- SmallWebRTC: the `/bot` request only waits for the connection, and the bot actually runs from a later `/api/offer` request. So the flow is stashed in `GLOBALS` next to the existing `pipecat_session_body` stash, picked up on the offer side, and cleared when the session ends.
- `tests/test_start_envelope.py` adds 14 tests. CHANGELOG and version bump to 0.1.30.

**Scope:** about 70 lines in the base image's request entry point, plus tests. It is on every session's start path, but a request without the header takes the old path unchanged. No schema or infra changes.

pipecat-base/app.py
```diff
@@ -286,6 +300,10 @@ async def run_bot(args: SessionArguments, transport_type: Optional[str] = None):
                 logger.info("Will wait for the webrtc_connection to be set!")
                 try:
                     GLOBALS["pipecat_session_body"] = args.body
+                    # The flow waits here with the body. This request only waits
+                    # for the connection; the bot runs from the later
+                    # /api/offer request, which never saw either of them.
+                    GLOBALS[_FLOW_CONFIG_KEY] = getattr(args, "flow_config", None)
                     await session_manager.wait_for_webrtc()
                 except TimeoutError as e:
                     logger.error(f"Timeout waiting for WebRTC connection: {e}")
```
cbrianhill: **1. A SmallWebRTC `/bot` that times out leaves the flow stashed** (`suggestion`)

The PR description says the flow "is cleared with [the body] when the session ends, so nothing carries into the next session on a reused pod." That is true when the WebRTC connection arrives. It is not true when it does not arrive. The clear only runs in the `finally` of the *offer-side* `run_bot` call. The `/bot` side re-raises the `TimeoutError` and returns before it reaches that `try`, so both stashes survive.

```python
# pipecat-base/app.py:297-332
session_manager = GLOBALS.get("session_manager")
if session_manager:
    if isinstance(args, PipecatSessionArguments) and transport_type == "webrtc":
        logger.info("Will wait for the webrtc_connection to be set!")
        try:
            GLOBALS["pipecat_session_body"] = args.body
            GLOBALS[_FLOW_CONFIG_KEY] = getattr(args, "flow_config", None)
            await session_manager.wait_for_webrtc()
        except TimeoutError as e:
            logger.error(f"Timeout waiting for WebRTC connection: {e}")
            raise                      # <-- stash left set
        return
    if isinstance(args, SmallWebRTCSessionArguments):
        # ...
        if not args.body:
            args.body = GLOBALS.get("pipecat_session_body")
        if getattr(args, "flow_config", None) is None:
            _attach_flow_config(args, GLOBALS.get(_FLOW_CONFIG_KEY))

try:
    await _run_bot_with_budget(args)
# ...
finally:
    # ...
    if session_manager:
        session_manager.complete_session()
        GLOBALS["pipecat_session_body"] = None
        GLOBALS[_FLOW_CONFIG_KEY] = None   # only reached by the offer-side call
```

`wait_for_webrtc` raises `TimeoutError` after 120 seconds if no offer arrives (`SmallWebRTCSessionManager` in daily-co/pipecat-cloud, `src/pipecatcloud/smallwebrtc/session_manager.py`).

What can read the stale value afterwards: the next webrtc `/bot` overwrites both stashes, so that path is fine. The WhatsApp webhook is not. It builds `SmallWebRTCSessionArguments` directly with no `/bot` before it and goes through the branch above:

```python
# pipecat-base/app.py:694-699
runner_args = SmallWebRTCSessionArguments(
    session_id=x_daily_session_id,
    webrtc_connection=connection,
    body=call,
)
background_tasks.add_task(run_bot, runner_args)
```

Its `flow_config` is always `None`, so it would pick up the previous session's flow. The body stash does less harm here because it is only used when `args.body` is empty. The flow is used whenever the args have none, which for WhatsApp is always. A late `/api/offer` from the timed-out client would also pick it up, but that one is at least the same caller's flow.

This needs both `SMALL_WEBRTC_SESSION` and `WHATSAPP` enabled on one agent, and a pod that serves another session after a failed `/bot`, so it is narrow. The body stash has the same gap today. Since the PR makes a specific claim about not carrying state across sessions, it is cheap to close:

```python
        except TimeoutError as e:
            logger.error(f"Timeout waiting for WebRTC connection: {e}")
            GLOBALS["pipecat_session_body"] = None
            GLOBALS[_FLOW_CONFIG_KEY] = None
            raise
```

This is safe on the success path. The offer side has already cleared both keys before `complete_session()` lets `wait_for_webrtc` return, so there is nothing left to lose there.

pipecat-base/app.py
```diff
@@ -349,6 +370,50 @@ async def livez():
...
+            was accepted.
+    """
+    if marker is None:
+        return body, None
+    if marker != _START_ENVELOPE_VERSION:
+        raise HTTPException(
+            status_code=400,
+            detail=f"unsupported start envelope version {marker!r}",
+        )
+    if not isinstance(body, dict) or "body" not in body or "flow_config" not in body:
```
cbrianhill: **2. A `null` or non-string `flow_config` passes the envelope check** (`nit`)

The check confirms the keys are present but not what `flow_config` holds:

```python
# pipecat-base/app.py:396-401
if not isinstance(body, dict) or "body" not in body or "flow_config" not in body:
    raise HTTPException(
        status_code=400,
        detail="start envelope needs both 'body' and 'flow_config'",
    )
return body["body"], body["flow_config"]
```

```python
# pipecat-base/app.py:413-414
if flow_config is not None:
    args.flow_config = flow_config
```

So `{"body": {...}, "flow_config": null}` with the header set is accepted, `_attach_flow_config` skips it, and the bot runs the flow its image ships with. That is the outcome the docstring says the 400 exists to prevent ("on the flow the image ships with while the caller believes theirs was accepted"). A JSON object instead of text also passes, and fails later inside the customer's `FlowConfig.from_yaml(...)`, after `/bot` has already accepted the session.

The platform cannot send either today, because sandbox#843 validates `flow_config` as `z.string().optional()` and only builds the envelope when it is set. So this is about the image side holding the contract on its own, which matters if anything else ever starts sending envelopes.

```python
flow_config = body["flow_config"]
if not isinstance(flow_config, str) or not flow_config:
    raise HTTPException(
        status_code=400,
        detail="start envelope 'flow_config' must be non-empty text",
    )
return body["body"], flow_config
```

pipecat-base/tests/test_start_envelope.py
```diff
@@ -0,0 +1,201 @@
...
+                return None
+
+        monkeypatch.setitem(app.GLOBALS, "session_manager", _Manager())
+        args = app.PipecatSessionArguments(session_id="sess-1241", body={"u": 1})
+        app._attach_flow_config(args, FLOW)
+
+        await app.run_bot(args, "webrtc")
+
+        assert app.GLOBALS["pipecat_session_body"] == {"u": 1}
+        assert app.GLOBALS[app._FLOW_CONFIG_KEY] == FLOW
```
cbrianhill: **3. The SmallWebRTC test covers the stash, not the pickup or the clear** (`suggestion`)

This test runs the `/bot` half and asserts the flow was stashed:

```python
# pipecat-base/tests/test_start_envelope.py:195-201
args = app.PipecatSessionArguments(session_id="sess-1241", body={"u": 1})
app._attach_flow_config(args, FLOW)

await app.run_bot(args, "webrtc")

assert app.GLOBALS["pipecat_session_body"] == {"u": 1}
assert app.GLOBALS[app._FLOW_CONFIG_KEY] == FLOW
```

The other two halves of what the PR describes are untested. First, the `/api/offer` side copying the stash onto `SmallWebRTCSessionArguments` (app.py:319-320). Second, the stash being cleared when that session ends (app.py:332). The clear is the part the description relies on for "nothing carries into the next session", and a regression there would not fail any test.

A second test could build `app.SmallWebRTCSessionArguments(session_id=..., webrtc_connection=None, body=None)` with `_FLOW_CONFIG_KEY` pre-set. It would stub `bot` (or `_run_bot_with_budget`) to capture the args, call `run_bot`, and assert that the captured `args.flow_config == FLOW` and that `GLOBALS[_FLOW_CONFIG_KEY] is None` afterwards. `SmallWebRTCSessionArguments` is only imported into `app` when `SMALL_WEBRTC_SESSION` is enabled, so the test may need to import it from `pipecatcloud.agent` directly.

Separately, the autouse fixture resets `_FLOW_CONFIG_KEY` and `_BUDGET_KEY` but not `pipecat_session_body`, which this test leaves set to `{"u": 1}` for any later test in the session.

cbrianhill (approved):
Checked what I could verify outside this diff:

- The post-construction assignment in `_attach_flow_config` (app.py:404-414) is safe on any pipecat-ai. `pipecatcloud.agent`'s `PipecatSessionArguments`, `DailySessionArguments` and `SmallWebRTCSessionArguments` are plain `@dataclass` subclasses of pipecat's `*RunnerArguments`, with no `slots=True` and no `frozen=True`, so `args.flow_config = ...` works whether or not the field exists on the class.
- The platform side (sandbox#843) matches this one. It sends `X-Pcc-Start-Envelope: 1`, validates `flow_config` as `z.string().optional()`, and gates on `FLOW_CONFIG_MIN_BASE_IMAGE_VERSION = '0.1.30'`, which is the version bumped in pyproject.toml here.
- `/ws` (app.py:457-481) never reads the envelope. That is consistent, because sandbox#843 refuses `flow_config` with `transport: "websocket"`.
- In the SmallWebRTC `finally` (app.py:329-332), `complete_session()` and the two `GLOBALS` clears run with no `await` between them, so the waiting `/bot` request cannot resume before the stash is cleared.
- The WhatsApp route (app.py:689-699) also builds `SmallWebRTCSessionArguments` and goes through the same `run_bot` branch, so it reads the flow stash too. After a normal session the stash is `None`, so this only matters on the path in finding 1.

The header marker, instead of inferring the envelope from the JSON shape, keeps a customer body that happens to contain `body` and `flow_config` keys intact. Both sides have a test for exactly that lookalike case.

Three findings, none blocking:

1. `suggestion`: a SmallWebRTC `/bot` that times out leaves the flow (and body) stashed, so a later WhatsApp session on the same pod could pick it up (app.py:309).
2. `nit`: an envelope with `"flow_config": null` or a non-string passes validation, and a `null` runs the image's own flow without error (app.py:396).
3. `suggestion`: the SmallWebRTC test checks the stash but not the pickup on the `/api/offer` side or the clear afterwards (tests/test_start_envelope.py:181-201).

Approved. The envelope handling and the version gate are correct against the platform PR. Finding 1 is an edge case inherited from how the body is already stashed, and the other two are hardening.

mattshep:
All three fixed in 444e68f.

**1.** Confirmed, and the chain holds: the timeout re-raises before the block that clears, and the WhatsApp route is the consumer that bites, since it builds its own SmallWebRTC arguments with no `/bot` before them and its flow is always absent. Both keys are cleared on that path now. I took the body stash with it — same two lines, and the claim this PR makes about state not crossing sessions covers both. The trade-off: a late offer from the timed-out client now runs the image's own flow instead of collecting its own. Same caller, on a session that already failed, which beats handing a flow to a different one.

**2.** Fixed. `flow_config` has to be non-empty text, so a `null` no longer passes the key check and then falls through to the shipped flow.

**3.** Added — the offer side collecting the flow, the clear that follows, and the timeout leaving nothing behind. The fixture gap was real too: it reset the flow key but not `pipecat_session_body`, so that test seeded whatever ran next.

68 tests, ruff clean.

mattshep:
One more in 6d04f64, since it is cheaper to decide before 0.1.30 publishes than after.

`flow_config` was assigned only when a session named one, so the attribute was absent entirely for every session that did not. The pipecat-ai an image is built against is its author's choice rather than ours, and the field arrives on `RunnerArguments` only from the release that adds it — so on an older one the read this change documents, `runner_args.flow_config`, raised `AttributeError` for most sessions. It is always set now, `None` when no flow was named, which on a newer pipecat-ai assigns what the field already defaults to. That decouples the base image from the pipecat-ai version instead of quietly requiring a recent one.

68 tests, ruff clean.

